PR #1, opened 2019-06-12
simple pecan app for marketplace with readme instructions for setup

fvukelic:


AUTHORS
```diff
@@ -0,0 +1,5 @@
+Sahil Tikale <sahil.tikale@gmail.com>
```
larsks: I would vote for getting rid of the `AUTHORS` file. This is what the git commit log is for. It's too easy for an `AUTHORS` file to fall out of sync as other people work on the project.

ChangeLog
```diff
@@ -0,0 +1,4 @@
+CHANGES
```
larsks: I would drop the `ChangeLog` for the same reason. While there are valid arguments for a manually curated ChangeLog, in general we should be writing commit messages such that `git log` by itself generally provides the same information.

flocx_market/templates/index.html
```diff
@@ -0,0 +1,34 @@
+<%inherit file="layout.html" />
```
larsks: We probably don't need this: this service implement a REST api, not a web UI.

flocx_market/templates/layout.html
```diff
@@ -0,0 +1,22 @@
+<html>
```
larsks: Same here: unnecessary for this project.

flocx_market/templates/error.html
```diff
@@ -0,0 +1,12 @@
+<%inherit file="layout.html" />
```
fvukelic: @larsks same for this one then I believe. Should I just remove the entire templates directory?

larsks: Probably, yeah.  Good catch.

flocx_market/templates/index.html
fvukelic: removed and made a new PR

flocx_market/templates/layout.html
fvukelic: removed and made a new PR

ChangeLog
fvukelic: done

AUTHORS
fvukelic: done

larsks:
@fvukelic I made a minor edit to a commit message, but otherwise this looks fine.

fvukelic:
great thanks!

PR #4, opened 2019-06-19
Implement common service code.

DanNiESh:
I followed the suggestion from Mainn and implemented the common service code. This pull request does not show the work on database models.
I imitated esi-leap code, created config files in conf folder, created service.py in common folder and also changed some files in api folder.
Now we can run the project with "sudo flocx-market-api", which is the entrypoint flocx_market.cmd.api:main (hopefully, if the pythonpath is correct)
I have some problems though:
1. When I run the project in my virtualenv with "sudo flocx-market-api", I get the error like "module not found". It's because it searches global python path for importing python modules but I installed the packages in the python inside my virtualenv. (I did export PYTHONPATH="/path/to/virtualenv/python, not working) However, if I run the project by "python flocx_market/cmd/api.py, it would work.
2. I cannot fully understand every part of the codes that I've made changes. Correct my work if some of the codes are redundant or can be better.


tzumainn:
Hi! This is looking good. I'm going to have a few random comments:

* You probably want to start using `git rebase` rather than `git pull`. The latter creates merge commits, while the former will re-apply your branch commits on top of any changes that might have happened on master.
* There are some merge conflicts that you'll need to resolve.
* It probably makes sense to squash some commits together - for example, the .gitignore updates.

I know this can be confusing, so if you want to walk through this stuff together sometime, just let me know!

larsks:
@tzumainn @DanNiESh  if you haven't seen them already, I wrote up [this article](https://blog.oddbit.com/post/2019-06-14-git-etiquette-commit-messages/) on making good pull requests and commit messages, and [this one](https://blog.oddbit.com/post/2019-06-17-avoid-rebase-hell-squashing-wi/) about squashing together a troublesome pr.

flocx_market/api/app.py
```diff
@@ -1,19 +1,46 @@
...
+        hooks=lambda: [ContextHook()],
+        debug=CONF.pecan.debug,
+        static_root=config.app.static_root if CONF.pecan.debug else None,
+        force_canonical=getattr(config.app, 'force_canonical', True),
+    )
+    # keystone authentication:
+    # if CONF.pecan.auth_enable:
+    #     app = auth_token.AuthProtocol(app, dict(CONF.keystone_authtoken))
+
     return app
```
tzumainn: These changes probably go in a separate PR, unless they're necessary with the service changes?

flocx_market/api/service.py
```diff
@@ -0,0 +1,50 @@
...
+        self.server.start()
+
+    def stop(self):
+        self.server.stop()
+
+    def wait(self):
+        self.server.wait()
+
+    def reset(self):
+        self.server.reset()
```
tzumainn: May as well add the newline

flocx_market/cmd/api.py
```diff
@@ -1,18 +1,32 @@
...
+    flocx_market_service.prepare_service(sys.argv)
+    # Build and start the WSGI app
+    launcher = service.ProcessLauncher(CONF, restart_method='mutate')
+    server = wsgi_service.WSGIService('flocx_market_api')
+    launcher.launch_service(server, workers=server.workers)
+    launcher.wait()
 
 if __name__ == '__main__':
-    main()
+    main()
```
tzumainn: newline here again

flocx_market/version.py
```diff
@@ -0,0 +1,15 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from pbr import version
+
+version_info = version.VersionInfo('flocx-market')
```
tzumainn: newline

flocx_market/api/app.py
tzumainn: Ah, I see - it's because of the conf changes!

DanNiESh: For service changes, I modified the configs into a conf folder. So I change the api.py accordingly. Probably some of the changes are not necessary, I'll fix it.

tzumainn:
Overall, I think the code looks good! I would encourage you to separate out the commits into smaller bits of work. Separating out the .gitignore commit is good; I'd go further by making separate commits for:

* the files in `conf/`
* the update to the code in `api/` to use the conf parameters

And other small, discrete updates that make sense by themselves.

PR #7, opened 2019-06-24
oslo configuration

ljmcgann:


tzumainn:
Thanks for the PR!

I think I'd start by separating out the oslo config updates from everything else - db, tests, etc. I'd suggest creating a new branch off of master, copying the relevant conf/* files, and perhaps making a new PR.

There are a few things missing: updates to setup.cfg, and additions to etc/flocx_marketplace, similar to what esi-leap has. Take a look, and let me know if you have any questions regarding what's going on there!

Don't forget to run pep8 before committing; that'll catch things like adding a newline at the end of the file, and other formatting issues like that.

This is a good start!

larsks:
...and I will remind you about https://blog.oddbit.com/post/2019-06-14-git-etiquette-commit-messages/, specifically: when you're making "fixing the previous commit" sorts of changes, don't make a new commit; just `commit --amend` and then update the PR.

larsks:
Also, in addition to what Mainn said about splitting things up, I would just get rid of `AUTHORS`  and `ChangeLog`; that information is better provided by `git log`.  While there are arguments for a curated changelog of some sort, there are better mechanisms for producing one (and I think we're better off without it right now).

flocx_market/cmd/api.py
```diff
@@ -1,16 +1,44 @@
...
+from flask import Flask, request
+from flask_restful import Api
+
+from resources.offer import Offer, OfferList
+from oslo_context import context
+import esi_leap.conf
+=======
 from wsgiref import simple_server
 from flocx_market.api import app
+>>>>>>> 5e3222c8... Flake 8 tests TG-57
```
larsks: It looks like you committed a file with some merge conflict markers. You'll want to watch out for this.

PR #8, opened 2019-06-25
Application rewritten to flask with a system test working that requires running db

fvukelic:
Tox and travis are failing.
Feedback is appreciated.


tzumainn:
Hi! This is impressive - nice work! My first high-level comment is that we probably want to reorganize some of the files to more closely match what we find in other OpenStack projects:

a) All the database stuff should be in db/ and db/migrations/
b) All the API stuff should be in api/
c) The unit test directory structure should match - so tests/db/, tests/api/, etc.

flocx_market/.env
```diff
@@ -0,0 +1,2 @@
+DATABASE_URI=mysql+pymysql://flocx_market:qwerty123@127.0.0.1:3307/flocx_market
+FLASK_APP=api.py
```
tzumainn: These files would be replaced by the oslo_config stuff that @ljmcgann  and @DanNiESh  are working on.

flocx_market/api.py
```diff
@@ -0,0 +1,25 @@
...
+
+from resources.offer import Offer, OfferList
+
+app = Flask(__name__)
+load_dotenv(".env")
+app.config['SQLALCHEMY_DATABASE_URI'] = os.environ.get('DATABASE_URI')
+app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+app.config['PROPAGATE_EXCEPTIONS'] = True
+api = Api(app)
+migrate = Migrate(app, db)
```
tzumainn: Starting up the API should not install the database tables (if I'm understanding this code correctly); they should be completely separate. You can see how Ironic or esi-leap does it, but perhaps it's fine for now, because the "OpenStack" way of doing things require the work that Danni and Leo are doing.

flocx_market/api.py
```diff
@@ -0,0 +1,25 @@
...
+api.add_resource(Offer, '/offer', '/offer/<string:marketplace_offer_id>')
+api.add_resource(OfferList, '/offers')
+
+if __name__ == '__main__':
+    db.init_app(app)
+
+    app.run(port=8080, debug=True)
```
tzumainn: I'm guessing this shouldn't be here, as it's cmd/app.py that starts up the app?

flocx_market/cmd/app.py
```diff
@@ -0,0 +1,13 @@
...
+from resources.offer import Offer, OfferList
+
+app = Flask(__name__)
+api = API(app)
+
+api.add_resource(Offer, '/offer/<string:id>')
+api.add_resource(OfferList, '/offers')
+
+if __name__ == '__main__':
+    app.run(debug = True)
```
tzumainn: This should probably be renamed api.py, and call a function in flocx_market/api.py. Lines 6-10 should be there, and not here.

flocx_market/resources/offer.py
```diff
@@ -0,0 +1,88 @@
...
+        if offer:
+            offer = OfferModel(marketplace_offer_id, **data)
+        offer.save_to_db()
+        return offer.json()
+
+
+class OfferList(Resource):
+    @classmethod
+    def get(cls):
+        return {"offers": [x.json() for x in OfferModel.query.all()]}
```
tzumainn: This probably belongs in the api/ directory. In the long run, I think we also want an objects/offer.py that is what connects to the database and performs any added business logic, so that api/offer.py only has to call methods there and is very thin. Probably fine for now though!

flocx_market/api.py
DanNiESh: Yes, Leo and I have moved api.py to cmd as app.py.

flocx_market/models/offer.py
```diff
@@ -0,0 +1,53 @@
...
+        self.creator_id = creator_id
+        self.marketplace_date_created = marketplace_date_created
+        self.status = status
+        self.server_id = server_id
+        self.start_time = start_time
+        self.end_time = end_time
+        self.server_config = server_config
+        self.cost = cost
+
+    def json(self):
```
larsks: This method should be renamed. It doesn't return a JSON representation of the data; it returns a dictionary.  So probably something like `as_dict`. You can probably simplify it like this:

```
def as_dict(self):
    return {col.name: getattr(self, col.name)
           for col in self.__table__.columns}
```

This way the method doesn't need to know about specific column names.

flocx_market/models/offer.py
```diff
@@ -0,0 +1,53 @@
+from db import db
```
larsks: There is no module named `db`:

```
>>> import flocx_market.models.offer
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lars/projects/flocx/flocx-market/flocx_market/models/offer.py", line 1, in <module>
    from db import db
ModuleNotFoundError: No module named 'db'
```

Please use fully qualified imports.  Instead of:

```
from db import db
```

Use:

```
from flocx_market.db import db
```

flocx_market/models/offer.py
```diff
@@ -0,0 +1,53 @@
...
+from db import sqlalchemy_jsonfield
```
larsks: This should just be:

```
import sqlalchemy_jsonfield
```

And you need to add `sqlalchemy_jsonfield` to `requirements.txt`.

flocx_market/api.py
fvukelic: hey, the database tables are only installed with migrations, starting it up won't initialize any tables. This is just configuration for the ORM. Does that answer your concern?

flocx_market/cmd/app.py
fvukelic: I accidentally left that file in there. I haven't used app.py for anything. Everything runs through api.py

flocx_market/api.py
fvukelic: @DanNiESh if we move this file to cmd/app.py, we should definitely leave the if __name__=='__main__' in there with the db init app and app.run

flocx_market/resources/offer.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument('cost',
+                        required=True,
+                        help="cost needed"
+                        )
+
+    @classmethod
+    def get(cls, marketplace_offer_id):
+        offer = OfferModel.find_by_id(marketplace_offer_id)
+        if offer:
+            return offer.json()
```
larsks: Why not return the `OfferModel` object here instead of converting it to a dictionary?

flocx_market/resources/offer.py
```diff
@@ -0,0 +1,88 @@
...
+        return {'message': 'Offer not found'}, 404
+
+    @classmethod
+    def post(cls):
```
larsks: Maybe this should be called `create`?

flocx_market/resources/offer.py
fvukelic: the objects you are referring to I believe can be found in the models, that is handling the db communication, and these resources are basically only the api endpoints responding to requests and calling the models that communicate with the db. 
I belive if we just rename the models to objects and move the resources to the api, that would match the openstack criteria. Do you agreee @tzumainn ?

flocx_market/resources/offer.py
```diff
@@ -0,0 +1,88 @@
...
+    @classmethod
+    def delete(cls, marketplace_offer_id):
+        offer = OfferModel.find_by_id(marketplace_offer_id)
+        if offer:
+            offer.delete_from_db()
+            return {'message': 'Offer deleted.'}
+        return {'message': 'Offer not found.'}, 404
+
+    @classmethod
+    def put(cls, marketplace_offer_id):
```
larsks: Maybe this should be called `update`?

flocx_market/api.py
tzumainn: It's possible I don't fully understand how Flask works! Does it require knowledge of the database?

flocx_market/models/offer.py
fvukelic: Agree for the renaming. 
If I try to do this, it throws a bug because the dateTime data types have to be casted to isoformat(). How do you recommend approaching it then?

flocx_market/models/offer.py
fvukelic: got it

flocx_market/resources/offer.py
fvukelic: the dateTime data type can't be deserialized, it will throw a bug

flocx_market/models/offer.py
fvukelic: sounds good, 
probably this spelling didn't work well "SQLAlchemy-JSONField>=0.8.0"

flocx_market/resources/offer.py
fvukelic: this is the api endpoint which calls the create/save_to_db()
I think since it extends resource, it has to follow the REST convention: get, post, put, etc.

flocx_market/resources/offer.py
fvukelic: this one should follow the REST names as well I'm pretty sure, but will look into it

flocx_market/.env
fvukelic: got it

flocx_market/api.py
larsks: Actually, you shouldn't have the `if name == '__main___'` block in there *anywhere*. You should be setting up a console_scripts entrypoint so that the service can be run from the command line.  See e.g. https://github.com/CCI-MOC/esi-leap/blob/master/setup.cfg#L15

flocx_market/models/offer.py
larsks: Just maintain a list of datetime columns (e.g. as a class attribute), I guess:

```
def as_dict(self):
    d = {}
    for col in self.__table__.columns:
        if col.name in self.datetime_columns:
            # ...perform conversion here...
        else:
            d[col.name] = getattr(self, col.name)

    return d
```

flocx_market/models/offer.py
larsks: That spelling *should* work, actually.  So maybe just fix the `import`.

flocx_market/resources/offer.py
larsks: Got it.

flocx_market/resources/offer.py
larsks: If that's the case, you can ignore this (and the following) comment.

flocx_market/api.py
fvukelic: the sqlalchemy track modifications if not specified will be set to TRUE and Flask will use it's own tracker which is poorer than SQL Alchemy's and that's why I set it to False.
For the propagate exceptions, I guess it's not that important since debug is already true:

"""PROPAGATE_EXCEPTIONS
Exceptions are re-raised rather than being handled by the app’s error handlers. If not set, this is implicitly true if TESTING or DEBUG is enabled.
Default: None"""


The orm does require some knowledge of the db such as how to access it

flocx_market/models/offer.py
fvukelic: sounds good

flocx_market/models/offer.py
fvukelic: ok, I'll look into it

flocx_market/api.py
fvukelic: okay then we need to take out db.init_app(app) into another file since that can only be called once, otherwise, as I read somewhere some kind of recursive imports will attempt to initialize the ORM every time app is used in another file and this can cause a lot of headaches. 
The name=='__main__' was also restricting it from calling db.init_app(app) when app is imported in another file. 
I'll play around with it.

flocx_market/models/offer.py
fvukelic: sounds good

flocx_market/api.py
tzumainn: Is there any way to separate the API from the database? Most OpenStack projects I've seen have three layers: database, business object, and REST API, with all the business logic encompassed in the business object. This allows for, say, a manager service to manipulate objects without going through the REST API (useful for expiring objects that have moved past a certain date).

DanNiESh: I'm also concerned about the db part inside api.py. I have nowhere to start with oslo db configuration. I guess we need a db folder with a db api.

fvukelic: yes we can separate them. However, the db piece will still need the Flask app to initialize the SQL Alchemy ORM @tzumainn

tzumainn: Is the "Flask app" the webapp that runs the REST API? If so, I think there's something problematic if the database layer requires that to work. Or I could be misunderstanding... ?

larsks: The flask app needs to initialize the ORM, but it shouldn't need to call any sort of database migration. The api code shouldn't know about alembic nor should it require the `flask_migrate` module. This should all be handled by a separate tool.

PR #10, opened 2019-06-25
[WIP] matcher - first commit

aliraza0337:
Hi, 
I created the matcher, it takes a list of expressions and ironic config in JSON format. It returns false if any of the expression doesn't hold true or it returns true if all the specs are found (as specified by the expressions). I also put some assertion instead of typecasting. not sure which one is better.

let me know what do you guys think about it? I am still working on it.

Best, 
Ali

### DIFF
--- a/flocx_market/tests/commit/test_commit_message.py
+++ b/flocx_market/tests/commit/test_commit_message.py
@@ -85,6 +85,19 @@ def test_commit_subject_length(commit, max_subject_length):
         '{} characters'.format(max_subject_length))
 
 
+def test_commit_line_length(commit):
+    '''Verify the length of the commit subject
+
+    This will fail if lines in the message are not
+    wrapped to < 75 characters.'''
+    message = commit.message.splitlines()
+    assert len(message) > 1
+
+    for line in message:
+        assert len(line) < 75, (
+            'Commit message body should be wrapped at 75 characters')
+
+
 def test_commit_has_reference(commit):
     '''Verify that a commit includes a bug/issue/task/etc reference'''
     has_gh_issue_refs = re.findall(r'#\d+', commit.message)


matcher/ironic_config.json
```diff
@@ -0,0 +1,473 @@
+{
```
larsks: I don't think this file should be part of the pull request. I think we will want some tests in this pr, but the tests don't need a full set of data like this. This is mostly useful for development, and it already lives in the flocx documentation repository (https://github.com/CCI-MOC/flocx/).

matcher/matcher.py
```diff
@@ -0,0 +1,109 @@
+import jmespath
+import json
+import re
+
+
+
+
+def apply_operator(j_val, op, val):
```
larsks: Calling the arguments `j_val` and `val` may be a bit confusing. Maybe just `val1` and `val2`?

matcher/matcher.py
```diff
@@ -0,0 +1,109 @@
...
+		if op == '<':
+			return j_val < val
+
+		if op == '<=':
+			return j_val <= val
+
+
+	# string operations
+	if op in all_str_ops:
+		assert isinstance(j_val,(str,unicode)), 'value is not str'
```
larsks: Same comment here re: the use of `assert`.

matcher/matcher.py
```diff
@@ -0,0 +1,109 @@
...
+			return j_val in val
+
+		if op == 'contains':
+			assert isinstance(j_val,(list,)), 'value is not str'
+			assert isinstance(val,(str,unicode)), 'value is not list'
+			return val in j_val
+
+
+
+def match_specs(match_expression, ironic_config):
```
larsks: This code doesn't really care of the data comes from ironic or somewhere else, so it's probably misleading to call this variable `ironic_config`.  Maybe just `data`?

matcher/matcher.py
```diff
@@ -0,0 +1,109 @@
...
+		if not apply_operator(j_val, op, val):
+			return False
+		
+	return True
+
+
+
+
+
+if __name__=="__main__":
```
larsks: I would get rid of this section and write a unit test to test the behavior of the code.

matcher/matcher.py
```diff
@@ -0,0 +1,109 @@
...
+	ironic_config = json.loads(open(file_name).read())
+	exps =[
+		    ["cpu_arch", "eq", "x86_64"],
+	  		["inventory.cpu.count", ">=", 16],
+	  		["inventory.cpu.flags", "contains", "vmx"], 
+	  		["length(inventory.disks[?\"rotational\" == `true`])", ">=", 2], 
+	  		["inventory.system_vendor.product_name", "matches", "PowerEdge M620"]
+	  	]
+	
+	print match_specs(exps,ironic_config)
```
larsks: This is not valid Python 3 (but if you move this into a unit test you won't be using `print` in any case, so this is mostly just an informational comment).

matcher/matcher.py
larsks: `unicode` is also not valid Python 3 (strings are unicode by default).

flocx_market/tests/unit/test_matcher.py
```diff
@@ -0,0 +1,130 @@
+from unittest import TestCase
```
larsks: You don't need `TestCase`.

flocx_market/tests/unit/test_matcher.py
```diff
@@ -0,0 +1,130 @@
...
+from matcher import * 
+import json
+
+data = json.loads(""" {
```
larsks: I would probably replace this giant chunk of data with a much smaller example dictionary. It should only contain what you need in order to run your tests.

flocx_market/tests/unit/test_matcher.py
larsks: You don't need this import (using pytest, we just write methods, rather than classses).

flocx_market/matcher.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # all numeric operations
+    if op in all_num_ops:
+
+        try:
+            val1 = float(val1)
+            val2 = float(val2)
+
+        except ValueError as verr:
+            print (verr.args)
```
larsks: It doesn't make any sense to call `print` in this code -- it's not going to be running on the console where anyone will see the output. If you need to log a message, using the `logging` module...but in this case, I think you should just remove the `try/except` block; it's okay to raise a `ValueError` if the function receives invalid values (and then add some tests that verify an exception is raised when you pass in invalid data).

flocx_market/matcher.py
```diff
@@ -0,0 +1,107 @@
...
+                return True
+            else:
+                return False
+
+    # list operators
+    if op in all_list_ops:
+
+        if op == 'in':
+            try:
+                val1 = str(val1)
```
larsks: I don't think we need explicit casting here; someone may conceivably be checking if a number is in a list, and this casting would make that test fail.

flocx_market/matcher.py
```diff
@@ -0,0 +1,107 @@
...
+                val2 = list(val2)
+            except ValueError as verr:
+                print (verr.args)
+                return False
+            return val1 in val2
+
+        if op == 'contains':
+            try:
+                val1 = list(val1)
```
larsks: Pretty much the same comment here as in the previous case.

test-requirements.txt
```diff
@@ -14,4 +14,4 @@ pytest>= 4.6.3
 tox>= 3.12.1 
 WebTest>=2.0.33
 flake8>= 3.7.7 
-
+jmespath>=0.9.4
```
larsks: This shouldn't be in `test-requirements.txt` (it's already in `requirements.txt`)

PR #11, opened 2019-06-25
Oslo Configuration

ljmcgann:
Oslo configuration is set up on MOC/flocx as opposed to on Filip's WIP code.


.gitignore
```diff
@@ -2,4 +2,8 @@
 .DS_Store
 build/
 flocx_market.egg-info/
-*/__pycache__/*
\ No newline at end of file
+*/__pycache__/*
+.idea/
+.tox/
+AUTHORS/
```
larsks: We probably don't need to list `AUTHORS` and `ChangeLog` in the `.gitignore` file (unless something is creating them for us automatically?). Also, since they have a trailing `/`, these entries would only match *directories* and would not match files.

requirements.txt
```diff
@@ -3,13 +3,32 @@
 # process, which may cause wedges in the gate later.
 pbr!=2.1.0,>=2.0.0 # Apache-2.0
 pycodestyle>=2.5.0
-Pecan>=1.3.3
+Click>=7.0
```
larsks: This appears to be introducing a bunch of requirements that aren't related to your configuration changes.

setup.cfg
```diff
@@ -8,6 +8,7 @@ packages =
     flocx_market
 
 [entry_points]
+oslo.config.opts =
+    flocx-market.conf = flocx_market.conf.opts:list_opts
 console_scripts =
-esi_leap.database.migration_backend = 
-	#sqlalchemy = esi_leap.db.sqlalchemy.migration
+    flocx-market-api = flocx_market.cmd.api:main
```
larsks: The console_scripts entrypoint does not appear to be related to the configuration changes.

flocx_market/conf/api.py
```diff
@@ -0,0 +1,23 @@
+from oslo_config import cfg
+
+
+opts = [
+    cfg.HostAddressOpt('host_ip',
+                       default='0.0.0.0'),
+    cfg.PortOpt('port',
+                default=8080),
```
tzumainn: 8080 is the default keystone port; let's change the default to something OpenStack doesn't use. For reference, here's a list of services and ports that are claimed: https://docs.openstack.org/openstack-helm/latest/devref/endpoints.html

flocx_market/conf/flask.py
```diff
@@ -0,0 +1,19 @@
+from oslo_config import cfg
+
+
+opts = [
+    cfg.StrOpt('SQLALCHEMY_DATABASE_URI',
+               default='mysql+pymysql://flocx_market:qwerty123@127.0.0.1:3306/flocx_market'),
```
tzumainn: I don't think we actually need this; including oslo.db in your config generator means that there'll be a [database] section that has a connection parameter.

Also, this sort of parameter probably should not have a default at all.

setup.cfg
```diff
@@ -8,6 +8,7 @@ packages =
     flocx_market
 
 [entry_points]
-console_scripts =
-esi_leap.database.migration_backend = 
+esi_leap.database.migration_backend =
 	#sqlalchemy = esi_leap.db.sqlalchemy.migration
+oslo.config.opts =
+    flocx-market.conf = flocx_market.conf.opts:list_opts
```
tzumainn: May as well put a newline here!

tzumainn:
Looks good - thanks!

tzumainn:
Resolved all of @larsks 's comments, so merging.

larsks:
@tzumainn I think that merge was premature: pull requests with code need to include unit tests for that code (and commit messages should be a little more filled out). @ljmcgann it would be great if you could prioritize writing some unit tests for this code before moving on to something else.  Thanks!

tzumainn:
@larsks I think it's fair to have slightly more descriptive commit messages. However, a conf tests are kind of difficult to write; are the tests checking to see that defaults are being set, or... ? Note that Ironic and Nova don't really have much in the way of tests for conf code either:

* https://github.com/openstack/nova/tree/master/nova/tests/unit/conf
* https://github.com/openstack/ironic/tree/master/ironic/tests/unit/conf

PR #17, opened 2019-06-27
oslo config cannot load config file in etc floder

DanNiESh:
This is a flask application.
I created a flocx_market service in common folder and a wsgi_service in api folder. The start_service.py is the script to launch the service. It seems like oslo cannot reach out to /etc/flocx-market/flocx-market.conf by default. 
The only way it works is to run the command "python flocx_market/cmd/start_service.py --config-file xx/xx/etc/flocx-market/flocx-market.conf"


DanNiESh:
Problem solved! Thanks

DanNiESh:
problem solved

PR #20, opened 2019-06-28
Add tests for Default oslo configurations

ljmcgann:
The default configurations created by oslo can be viewed through the CONF object. Modifications to these variables in /etc/flocx-market/flocx-market.conf can also be seen in the CONF object

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ flocx_market.egg-info/
 AUTHORS
 ChangeLog
 etc/flocx-market/flocx-market.conf.sample
+*.pyc
--- a/flocx_market/tests/unit/conf/test_conf.py
+++ b/flocx_market/tests/unit/conf/test_conf.py
@@ -1,0 +1,25 @@
+import flocx_market.conf as conf
+import socket
+
+CONF = conf.CONF
+
+
+def test_api_defaults():
+
+    assert(CONF.api.host_ip == "0.0.0.0")
+    assert(CONF.api.port == 8081)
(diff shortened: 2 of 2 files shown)


flocx_market/tests/unit/test_auth.py
```diff
@@ -0,0 +1,31 @@
...
+
+    assert(CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS is False)
+    assert(CONF.flask.PROPAGATE_EXCEPTIONS is False)
+
+
+def test_netconf_defaults():
+
+    assert(CONF.host == socket.gethostname())
+
+flocx_market_config.prepare_service()
```
tzumainn: You pretty much explicitly do *not* want this. Your configuration defaults will be instantiated when you import flocx_market.conf, since flocx_market/conf/__init__.py handles that. All this does is tell the tests to *also* look at /etc/flocx-market/flocx-market.conf, which will cause the test to fail if defaults are overwritten.

flocx_market/tests/unit/test_auth.py
```diff
@@ -0,0 +1,31 @@
...
+test_api_defaults()
+test_flask_defaults()
+test_netconf_defaults()
```
tzumainn: These three lines are also extraneous; you don't need them to have the tests run when you run 'tox'

tzumainn:
Oh! And you may want to change your commit message title to something more description - something about oslo configuration tests.

PR #23, opened 2019-06-28
Limit pep8 check to flocx_market directory

### LINKED ISSUE #22: the pep8 check runs erroneously on *all* python files, even those not part of the repository
If you have a virtual environment in your working directory, running the `pep8` check (`tox -e pep8`) will check not just our files but *all* files in your virtual environment. This can take a long time and will produce erroneous results.

larsks:
Limit the pep8 check to the flocx_market directory, to avoid
checking Python files that are not part of the repository. We also add
{posargs} to most tox checks to allow additional configuration from
the command line.

Closes #22

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,11 +16,11 @@ deps = -r{toxinidir}/requirements.txt
 
 [testenv:pep8]
 commands =
-        ./flakeight.sh
+        ./flakeight.sh {posargs:flocx_market}
 
 [testenv:py37]
 commands = 
-        pytest flocx_market/tests/unit/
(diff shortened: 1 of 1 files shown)


larsks:
@aliraza0337 this commit fixes the problem you were having in which the `pep8` check was checking files in your virtual environment.

tzumainn:
works with my ad-hoc test, but I'll wait to make sure that it works for @aliraza0337 as well!

aliraza0337:
Yes, It worked fine with mine too and didn't test files in my virtual environment.

PR #25, opened 2019-06-28
Hooks for default config file and business logic objects

ljmcgann:
Created service.py which hooks to oslo the configuration options from
the config file located at /etc/flocx-market/flocx-market.conf to make
them accessible by the CONF object. Additionally, added the version and
objects hooks, so oslo cna be aware of the version through pbr and the
business logic objects we will eventually create in the
flocx_market/objects directory

### DIFF
--- a/flocx_market/common/service.py
+++ b/flocx_market/common/service.py
@@ -1,0 +1,26 @@
+from oslo_db import options as db_options
+from oslo_log import log
+from oslo_service import service
+
+import flocx_market.conf
+from flocx_market import objects
+from flocx_market import version
+
+CONF = flocx_market.conf.CONF
+
+
+def prepare_service(argv=None, default_config_files=None):
+    argv = [] if argv is None else argv
+    log.register_options(CONF)
+    CONF(argv[1:],
+         project='flocx-market',
+         version=version.version_info.release_string(),
+         default_config_files=default_config_files)
+
+    db_options.set_defaults(CONF)
+    log.setup(CONF, 'flocx-market')
+    objects.register_all()
+
+
+def process_launcher():
+    return service.ProcessLauncher(CONF, restart_method='mutate')
--- a/flocx_market/conf/api.py
+++ b/flocx_market/conf/api.py
@@ -21,3 +21,4 @@ api_group = cfg.OptGroup(
 
 def register_opts(conf):
     conf.register_opts(opts, group=api_group)
+    conf.register_cli_opts(opts, group=api_group)
--- a/flocx_market/conf/flask.py
+++ b/flocx_market/conf/flask.py
@@ -15,3 +15,4 @@ flask_group = cfg.OptGroup(
 
 def register_opts(conf):
     conf.register_opts(opts, group=flask_group)
+    conf.register_cli_opts(opts, group=flask_group)
--- a/flocx_market/conf/netconf.py
+++ b/flocx_market/conf/netconf.py
@@ -11,3 +11,4 @@ opts = [
 
 def register_opts(conf):
     conf.register_opts(opts)
+    conf.register_cli_opts(opts)
--- a/flocx_market/objects/__init__.py
+++ b/flocx_market/objects/__init__.py
@@ -1,0 +1,6 @@
+# put import statements for all objects in the objects directory in register_all
+# Called as a hook in common/services.py
+
+
+def register_all():
+    pass
--- a/flocx_market/tests/unit/common/test_service.py
+++ b/flocx_market/tests/unit/common/test_service.py
@@ -1,0 +1,66 @@
+import flocx_market.conf as conf
+import socket
+import flocx_market.common.service as flocx_market_service
+import pytest
+import unittest.mock as mock
+
+CONF = conf.CONF
+
+
+@mock.patch('oslo_config.cfg.find_config_files')
+def test_service_defaults(find_config_files):
+
+    CONF.clear()
+    flocx_market_service.prepare_service()
+    assert(CONF.api.host_ip == "0.0.0.0")
+    assert(CONF.api.port == 8081)
+    assert(CONF.api.max_limit == 1000)
+    assert(CONF.api.public_endpoint is None)
+    assert(CONF.api.api_workers is None)
+    assert(CONF.api.enable_ssl_api is False)
+    assert(CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS is False)
+    assert(CONF.flask.PROPAGATE_EXCEPTIONS is False)
+    assert(CONF.host == socket.gethostname())
+
+
+@mock.patch('oslo_config.cfg.find_config_files')
+def test_service_cli_valid(find_config_files):
+
+    CONF.clear()
+    args = ['dummy', '--api-port', '8085', '--api-host_ip', '0.0.0.1',
+            '--api-public_endpoint', '/index', '--api-api_workers', '5', '--api-enable_ssl',
+            '--flask-noSQLALCHEMY_TRACK_MODIFICATIONS', '--flask-PROPAGATE_EXCEPTIONS', '--host', 'imhost']
+
+    flocx_market_service.prepare_service(argv=args)
+    assert(CONF.api.port == int(args[args.index('--api-port') + 1]))
+    assert(CONF.api.host_ip == str(args[args.index('--api-host_ip') + 1]))
+    assert(CONF.api.public_endpoint == str(args[args.index('--api-public_endpoint') + 1]))
+    assert(CONF.api.api_workers == int(args[args.index('--api-api_workers') + 1]))
+    assert(CONF.api.enable_ssl_api)
+
+    assert(CONF.api.max_limit == 1000)
+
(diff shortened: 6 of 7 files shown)


tzumainn:
I'm okay with this because I'm not sure there are useful unit tests for this code, but I'll defer to @larsks 's opinion on that!

larsks:
I think we do need unit tests there, because in addition to verifying that things work as intended that helps protect against things like someone accidentally introducing a typo in the future, etc. I figure a test for prepare_service() called with default parameters, a test for prepare_service() called with a valid command line option (verifying that it sets the value in CONF as expected), and a test for calling prepare_service() with an invalid command line option (verifying that it fails as expected).

flocx_market/tests/unit/common/test_prepareservice.py
```diff
@@ -0,0 +1,58 @@
...
+    assert(CONF.api.port == 8081)
+    assert(CONF.api.max_limit == 1000)
+    assert(CONF.api.public_endpoint is None)
+    assert(CONF.api.api_workers is None)
+    assert(CONF.api.enable_ssl_api is False)
+    assert(CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS is False)
+    assert(CONF.flask.PROPAGATE_EXCEPTIONS is False)
+    assert(CONF.host == socket.gethostname())
+
+    CONF.clear()
```
larsks: Calling `CONF.clear()` at the end of the test isn't going to work reliably, because in the event of a failure in any of the above `assert` statements, your `CONF.clear()` will not get called (which will in turn cause subsequent tests to fail).  Can you call `CONF.clear()` at the *beginning* of each test?

flocx_market/tests/unit/common/test_prepareservice.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_service_cli_valid():
+
+    args = ['dummy', '--api-port', '8085', '--api-host_ip', '0.0.0.1',
```
larsks: I think that just testing a single option would be sufficient.

flocx_market/tests/unit/conf/test_conf.py
```diff
@@ -1,9 +1,11 @@
 import flocx_market.conf as conf
 import socket
+import pytest
 
 CONF = conf.CONF
 
 
+@pytest.fixture()
```
larsks: I think you may have misunderstood the use of the `pytest.fixture` decorator.  These aren't fixtures, and you can safely remove the decorator from all the tests.

flocx_market/tests/unit/common/test_prepareservice.py
```diff
@@ -0,0 +1,58 @@
+import flocx_market.conf as conf
+import socket
+import flocx_market.common.service as flocx_market_service
+import pytest
+
+CONF = conf.CONF
+
+
+def test_service_defaults():
+    flocx_market_service.prepare_service()
```
tzumainn: Calling this as-is will not always get the result you want; if the person running the test has a /etc/flocx-market/flocx-market.conf configuration file with non-default values, then this test will use those values and fail.

flocx_market/tests/unit/common/test_prepareservice.py
ljmcgann: I don't believe calling clear at the beginning will work as any call to prepare_service will modify the state of CONF in some way.

flocx_market/tests/unit/conf/test_conf.py
ljmcgann: I forgot to remove these. My bad

flocx_market/tests/unit/common/test_prepareservice.py
larsks: I think it *has* to be called at the beginning of the function, because as I pointed out, calling it at the end is problematic. It's okay that prepare_service modifies it; the goal is simply to clear out the changes made by a *prior* call to prepare_service.

flocx_market/common/service.py
```diff
@@ -0,0 +1,29 @@
...
+def prepare_service(argv=None, default_config_files=None):
+    argv = [] if argv is None else argv
+    log.register_options(CONF)
+    try:
+        CONF(argv[1:],
+             project='flocx-market',
+             version=version.version_info.release_string(),
+             default_config_files=default_config_files)
+    except SystemExit:
+        raise SystemExit
```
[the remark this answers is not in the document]
ljmcgann: I fumbled around with the exceptions for a while. And its probably because I don't understand python exception handling well enough but not having this try catch block there in some capacity wasnt working in the test function and Sahil had to help me to get this working in the way that it is

flocx_market/tests/unit/common/test_prepareservice.py
larsks: (You could handle this by using a pytest fixture that would call `CONF.clear` before each test; that would make the implementation a little cleaner but the effect would be the same).

flocx_market/tests/unit/common/test_prepareservice.py
ljmcgann: I knew this but I wasn't sure how to test defaults otherwise. Do you have any recommendations? Or do you think the test is redundant with tests/conf/test_conf anyways?

tzumainn:
Thinking about it, I also don't think there's a reason to expose every option to the CLI. Maybe just the API ones?

flocx_market/common/service.py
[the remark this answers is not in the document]
larsks: You should remove this `try/except` block.  Can you explain in more detail what sort of problem you were having with the tests?

flocx_market/tests/unit/common/test_prepareservice.py
larsks: Can you explicitly point it at a non-default config file (like `/dev/null`)?

flocx_market/tests/unit/common/test_prepareservice.py
ljmcgann: I was testing to make sure erroneous arguments as well as incorrect argument types could not be passed into CONF, but if you feel this is unneeded I will remove it

flocx_market/tests/unit/common/test_prepareservice.py
```diff
@@ -0,0 +1,58 @@
...
+    args = ['dummy', '--badargs', '--api-port', 'notInt']
+
+    with pytest.raises(SystemExit):
+        flocx_market_service.prepare_service(argv=args)
+
+    CONF.clear()
+    args.remove('--badargs')
+    with pytest.raises(SystemExit):
+        flocx_market_service.prepare_service(argv=args)
+
```
larsks: You would probably want to verify that SystemExit was raised with an exitcode of 2, and maybe split this into two separate tests:



```
def test_service_cli_invalid_option():
    CONF.clear()
    with pytest.raises(SystemExit) as excinfo:
        flocx_market_service.prepare_service(
            argv=['dummy', '--badarg'])

    assert excinfo.value.code == 2

def test_service_cli_invalid_value():
    CONF.clear()
    with pytest.raises(SystemExit):
        flocx_market_service.prepare_service(
            argv=['dummy', '--api-port', 'notInt'])

    assert excinfo.value.code == 2
```

flocx_market/common/service.py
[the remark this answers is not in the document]
ljmcgann: Hmm. I'm not sure what I did differently this time around but now I don't need the try catch. I guess I was looking for the wrong exception on the testers side or something. My bad

flocx_market/tests/unit/common/test_prepareservice.py
larsks: I agree that we need to test for incorrect arguments and invalid types, but that's not what this test is doing: it is checking that valid options work correctly. You address invalid options and types in subsequent tests. That said, I don't think you need to change this: it works fine, I think it's just longer than it needs to be. I'm happy with this one either way.

flocx_market/tests/unit/common/test_prepareservice.py
tzumainn: Tested this out a bit, and adding an import and changing the test a bit works:

```
import unittest.mock as mock

@mock.patch('oslo_config.cfg.find_config_files')
def test_service_defaults(find_config_files):
      .
      .
```

I'd recommend doing that for all of these tests.

tzumainn: Just for clarity, oslo_config.cfg.find_config_files is what searches for config files in certain default locations; mocking that method guarantees it will always return nothing, forcing the tests to rely on the defaults.

flocx_market/tests/unit/common/test_prepareservice.py
ljmcgann: The exit code for the second test is None. The oslo_config code does not set a value for it. I will just leave it as assert is None

tzumainn:
I'm good with this now - thanks for the changes, @ljmcgann!

@larsks, any reservations?

larsks:
Looks good to me!

PR #26, opened 2019-06-29
Rewrote web service using flask. Implemented Marketplace API.

fvukelic:
Implemented offers, contracts,bids and restructured the project directory structure to meet OpenStack standards.

- TG-67
- TG-68
- TG-69
- TG-71
- TG-72
- TG-73
- TG-89
- TG-102


test-requirements.txt
```diff
@@ -3,15 +3,39 @@
...
+oslo.service!=1.28.1,>=1.24.0 # Apache-2.0
+oslo.upgradecheck>=0.1.0 # Apache-2.0
+oslo.utils>=3.33.0 # Apache-2.0
+oslo.versionedobjects>=1.31.2 # Apache-2.0
+jmespath>=0.9.4
 pytest>= 4.6.3
-tox>= 3.12.1 
 WebTest>=2.0.33
-flake8>= 3.7.7 
-GitPython>=2.1.11
```
tzumainn: I think the commit message test needs this.

.gitignore
```diff
@@ -1,11 +1,17 @@
...
 .DS_Store
+venv3.7/
 build/
 flocx_market.egg-info/
 */__pycache__/*
 .idea/
 .tox/
 AUTHORS
 ChangeLog
+<<<<<<< HEAD
```
tzumainn: Looks like a rebase conflict?

flocx_market/.env
```diff
@@ -0,0 +1,2 @@
+DATABASE_URI=mysql+pymysql://flocx_market:qwerty123@127.0.0.1:3307/flocx_market
```
tzumainn: It'd be much much better to use oslo config to get this; @ljmcgann and @DanNiESh should be able to give you pointers on how!

flocx_market/.env
```diff
@@ -0,0 +1,2 @@
...
+FLASK_APP=api.py
```
tzumainn: Not entirely sure what this is used for... ? In any case, starting up the API should be done through an entrypoint in setup.cfg - @ljmcgann and @DanNiESh should be able to help with that.

flocx_market/.env.example
```diff
@@ -0,0 +1 @@
+DATABASE_URI=
```
tzumainn: If we use oslo config, this file can be cut.

flocx_market/api/app.py
```diff
@@ -1,20 +1,26 @@
...
-    app_conf = dict(config.app)
-    app = pecan.make_app(
-        app_conf.pop('root'),
-        logging=getattr(config, 'logging', {}),
-        **app_conf)
+def create_app(app_name):
+
+    app = Flask(app_name)
+    load_dotenv(".env")
+    app.config['SQLALCHEMY_DATABASE_URI'] = os.environ.get('DATABASE_URI')
```
tzumainn: This is where we'd fetch the database connection string from the config file using oslo config code.

flocx_market/cmd/app.py
```diff
@@ -0,0 +1,11 @@
...
+from flocx_market.db.orm import orm
+from flask_migrate import Migrate
+
+application = app.create_app(app_name='flocx-market')
+migrate = Migrate(application, orm)
+
+if __name__ == '__main__':
+
+    orm.init_app(application)
+    application.run(port=8080, debug=True)
```
tzumainn: This should use the common service code that initializes things like the config file, the log, etc; I think @DanNiESh or @ljmcgann is working on that.

test-requirements.txt
larsks: `flake8` is required by the pep8 tests, and `GitPython` is required by the commit message check.

fvukelic: yes, I accidentally removed those, I'll put them back

flocx_market/.env
DanNiESh: Sure. We can use oslo config to replace the environment variables and hardcode lines @fvukelic

flocx_market/.env
larsks: The `.env` file is used by `docker-compose`; it should not be part of this commit.

.gitignore
fvukelic: yes that's a mistake

flocx_market/cmd/app.py
DanNiESh: The common service code works. We can have a discuss and put these codes into your application. @fvukelic

.gitignore
```diff
@@ -1,11 +1,17 @@
 .eggs/
 .DS_Store
+venv3.7/
```
larsks: I would probably not include this in the `.gitignore` file: it's not a standard name nor is it created by anything automatically; we don't want this file to grow to include everyone's individual preferences for virtual environment names.  You can use the file `.git/info/exclude` as a local `.gitignore` that isn't part of the repository.

flakeight.sh
```diff
@@ -7,13 +7,5 @@
...
-    files=$(git diff --name-only HEAD~1 | tr '\n' ' ')
-    echo "Running flake8 on ${files}"
-    diff -u --from-file /dev/null ${files} | flake8 --diff "$@"
-else
-    echo "Running flake8 on all files"
-    exec python3 -m flake8 "$@"
-fi
\ No newline at end of file
+echo "Running flake8 on flocx_market"
+exec python3 -m flake8 "flocx_market"
```
larsks: I'm not sure I understand the purpose of this change; the existing script was useful because it permitted running flake8 on only the files in the current commit.  For targeting the `flocx_market` directory specifically by default, see https://github.com/CCI-MOC/flocx-market/pull/23.

I would prefer to leave this file unmodified in this pull request.

flocx_market/.env
larsks: This is not directly related to the use of oslo_config (or not); it's a docker-compose artifact. It should not be in this pull request.

flocx_market/api/app.py
```diff
@@ -1,20 +1,26 @@
...
+    app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+    app.config['PROPAGATE_EXCEPTIONS'] = True
+    api = Api(app)
+    api.add_resource(Offer, '/offer', '/offers/<string:marketplace_offer_id>')
+    api.add_resource(OfferList, '/offers')
```
larsks: There is no `/offers` endpoint in the API (see https://github.com/CCI-MOC/flocx/blob/master/docs/api_marketplace.yaml#L37 and following).

flocx_market/api/app.py
```diff
@@ -1,20 +1,26 @@
...
+    api.add_resource(Bid, '/bid', '/bids/<string:marketplace_bid_id>')
+    api.add_resource(BidList, '/bids')
+    api.add_resource(Contract, '/contract', '/contracts/<string:contract_id>')
+    api.add_resource(ContractList, '/contracts')
```
larsks: ...there is also no `/bids` and no `/contracts`.

flocx_market/db/sqlalchemy/bid_api.py
```diff
@@ -0,0 +1,38 @@
+from flocx_market.db.orm import orm
+from flocx_market.db.sqlalchemy.bid_model import BidModel
+from uuid import uuid4
+
+
+class BidApid(BidModel):
```
larsks: Why is this `BidApid` and not just `BidApi`?

flocx_market/db/sqlalchemy/contract_api.py
```diff
@@ -0,0 +1,48 @@
...
+        self.cost = cost
+
+    def as_dict(self):
+        d = {}
+        for col in self.__table__.columns:
+            if col.name in ['time_created', 'start_time', 'end_time']:
+                d[col.name] = getattr(self, col.name).isoformat()
+            else:
+                d[col.name] = getattr(self, col.name)
+        offer_model_objects = [offer for offer in self.offers.all()]
```
larsks: For performance reasons, I don't think we want to expand offer ids into offer objects here (or at least, not by default). It is typical for REST APIs to either (a) not expand referenced objects by default (leaving it up to the client) or (b) only do so when explicitly requested.

flocx_market/tests/base_test.py
```diff
@@ -0,0 +1,36 @@
+from unittest import TestCase
+from flocx_market.cmd.app import application
+from flocx_market.db.orm import orm
+
+
+class BaseTest(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # ensure db exists
+        application.config['SQLALCHEMY_DATABASE_URI'] = 'postgresql://gpdwpyrw:UCzgXashsqnX99ASMK2SxZK8d4tbEUa6@raja.db.elephantsql.com:5432/gpdwpyrw'
```
larsks: Integration testing is a great idea!  But I don't think we should have any tests that rely on an external service. A typical way of handling integration tests is to have the CI environment set up whatever services are necessary for testing, and then pass in the necessary configuration to the tests.  I would suggest simply removing everything but the unit tests from this pull request, and then implement integration tests separately.

test-requirements.txt
```diff
@@ -3,15 +3,39 @@
...
+oslo.log>=3.36.0 # Apache-2.0
+oslo.messaging>=5.29.0 # Apache-2.0
+oslo.middleware>=3.31.0 # Apache-2.0
+oslo.policy>=1.30.0 # Apache-2.0
+oslo.serialization!=2.19.1,>=2.18.0 # Apache-2.0
+oslo.service!=1.28.1,>=1.24.0 # Apache-2.0
+oslo.upgradecheck>=0.1.0 # Apache-2.0
+oslo.utils>=3.33.0 # Apache-2.0
+oslo.versionedobjects>=1.31.2 # Apache-2.0
+jmespath>=0.9.4
```
larsks: You shouldn't be duplicating items here that are already included in `requirements.txt`.  `test-requirements.txt` is only for packages required to run the tests.

flakeight.sh
fvukelic: Ok I'll just move the virtual env file elsewhere so flake8 doesn't throw a bunch of errors for it

flocx_market/db/sqlalchemy/bid_api.py
fvukelic: mistake, will fix it

flakeight.sh
larsks: If you rebase on `master` then flake8 will only run on the `flocx_market` directory by default (#23 has merged).

test-requirements.txt
fvukelic: I just realized, it was above there already, I just removed the duplicates

PR #30, opened 2019-07-02
 Rewrote web service using Flask instead of Pecan. Implemented offers API.

fvukelic:
Implemented offers endpoints, and restructured the project directory structure to meet OpenStack standards.
- TG-86
- TG-89
- TG-102


flocx_market/tests/base_test.py
```diff
@@ -0,0 +1,36 @@
+from unittest import TestCase
+from flocx_market.cmd.app import application
+from flocx_market.db.orm import orm
+
+
+class BaseTest(TestCase):
```
tzumainn: Not all tests will involve the database, so I might have the BaseTest just set the context, and have something like BaseDBTest that extends BaseTest and which sets up the database.

fvukelic: Hey, I only use BaseTest for integration and System tests which usually involve some kind of crud operation. For unit tests, I use UnitBaseTest which doesn't involve the DB. 
What do you think about this approach @tzumainn ?

tzumainn: Ah, okay - that's fair - thanks!

flocx_market/tests/integration/models/test_offer.py
```diff
@@ -0,0 +1,21 @@
...
+
+class TestOffer(BaseTest):
+
+    def test_crud(self):
+        with self.app_context():
+            offer = OfferApi('b9752cc0-9bed-4f1c-8917-12ade7a6fdbe',
+                             'Alice1992',
+                             now, 'available', 'Dell380', now,
+                             now,
+                             '{"new attribute XYZ": "This is just a sample list of free-form attributes used for describing a server.","cpu_type": "Intel Xeon","cores": 16,"ram_gb": 512,"storage_type": "samsung SSD", "storage_size_gb": 204}',
```
larsks: This looks like a problem: The offer should accept a *dictionary* for the `server_config` field. The serialization to JSON for storage in the datbase (and un-serialization when a value is loaded) is an implementation detail that should not be visible to calling code.

flocx_market/api/app.py
```diff
@@ -1,25 +1,14 @@
-import os
 from flask import Flask
-from flask import Blueprint
-from flask import jsonify
-import flocx_market.conf
-CONF = flocx_market.conf.CONF
```
tzumainn: You're almost definitely going to need this.

flocx_market/api/app.py
```diff
@@ -1,25 +1,14 @@
...
-    return jsonify(version)
+from flocx_market.api.offer import Offer, OfferList
 
 
 def create_app(app_name):
     app = Flask(app_name)
-    app.register_blueprint(flocx_market_api_bp)
+    app.config['SQLALCHEMY_DATABASE_URI'] = 'mysql+pymysql://flocx_market:qwerty123@127.0.0.1:3307/flocx_market'
+    app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+    app.config['PROPAGATE_EXCEPTIONS'] = True
```
tzumainn: Two comments:

* These values should be retrieved using oslo config - CONF.database.connection, and CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS, etc - ask @DanNiESh or @ljmcgann if you want help with that

* The database configuration should not be done here, but somewhere in db/sqlalchemy/. Remember - the REST API is not the only route into working with these objects, and we don't want to duplicate this in multiple places. Take a look at https://github.com/openstack/ironic/blob/master/ironic/db/sqlalchemy/models.py for reference.

flocx_market/cmd/app.py
```diff
@@ -0,0 +1,11 @@
+from flocx_market.api import app
+from flocx_market.db.orm import orm
+from flask_migrate import Migrate
+
+application = app.create_app(app_name='flocx-market')
+migrate = Migrate(application, orm)
```
tzumainn: This should be done in a separate script, something like cmd/dbsync.py. Something like https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/cmd/dbsync.py.  @DanNiESh can help if you have questions on that!

flocx_market/cmd/app.py
```diff
@@ -0,0 +1,11 @@
...
+
+if __name__ == '__main__':
+
+    orm.init_app(application)
+    application.run(port=8080, debug=True)
```
tzumainn: I'm not sure why we want this over cmd/api.py. I'd keep cmd/api.py for starting up the api, and make any changes needed there.

flocx_market/db/sqlalchemy/offer_model.py
```diff
@@ -0,0 +1,16 @@
...
+    __tablename__ = 'offers'
+    marketplace_offer_id = orm.Column(orm.String(64), primary_key=True, autoincrement=False)
+    provider_id = orm.Column(orm.String(64), nullable=False)
+    creator_id = orm.Column(orm.String(64), nullable=False)
+    marketplace_date_created = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    status = orm.Column(orm.String(15), nullable=False, default='available')
+    server_id = orm.Column(orm.String(64), nullable=False, unique=True)
+    start_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    end_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    server_config = orm.Column(sqlalchemy_jsonfield.JSONField(enforce_string=True, enforce_unicode=False), nullable=False)
```
tzumainn: I've had more success with this sort of field as an oslo JsonEncodedDict, as in
https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/db/sqlalchemy/models.py#L57

flocx_market/cmd/app.py
fvukelic: This is the only thing I have for starting up the api. SQLalchemy needs to be initialized when the app/api starts and application.run is starting up the api

tzumainn: Either way, we should be updated cmd/api.py instead.

fvukelic: I'm a bit unclear on what are you referring to with updating cmd/api.py. Should I just rename cmd/app.py to api.py?

tzumainn: Oh, so there's merged code from @DanNiESh that already starts up a flask app, using the code in flocx_market/cmd/api.py. We should continue to use that!

DanNiESh: > I'm a bit unclear on what are you referring to with updating cmd/api.py. Should I just rename cmd/app.py to api.py?

Please fetch the upstream, and use cmd: flocx-market-api to start the project. This cmd will run cmd/api.py.

tzumainn:
I think lars removed flakeight.sh and moved it elsewhere in a recently moved PR; if you remove it from this PR, I think the conflicts will be fixed!

flocx_market/db/sqlalchemy/offer_api.py
```diff
@@ -0,0 +1,41 @@
...
+        self.server_config = json.dumps(server_config)
+        self.cost = cost
+
+    def as_dict(self):
+        d = {}
+        for col in self.__table__.columns:
+            if col.name in ['marketplace_date_created', 'start_time', 'end_time']:
+                d[col.name] = getattr(self, col.name).isoformat()
+            elif col.name in ['server_config']:
+                d[col.name] = json.loads(getattr(self, col.name))
```
larsks: I would really like to see this fixed before we merge these changes: the data model should return a Python dictionary, not a JSON string. The format in which the data is stored in the database should be transparent to client code. If we want to stop using JSON and use something else, it should *only* require changes to the data model and nowhere else. Any error handling relating to JSON serialization/deserialization should be handled by the data model, rather than requiring client code to handle that.

flocx_market/tests/unit/unit_base_test.py
```diff
@@ -0,0 +1,6 @@
+from flocx_market.cmd.api import application
```
tzumainn: This import isn't necessary

flocx_market/cmd/api.py
```diff
@@ -11,19 +11,35 @@
...
+from flask_migrate import Migrate
+from flocx_market.db.orm import orm
+
+
 CONF = flocx_market.conf.CONF
+flocx_market_service.prepare_service(sys.argv)
+
+
+application = app.create_app(app_name='flocx-market')
+migrate = Migrate(application, orm)
```
tzumainn: these lines - 26-30 - should be *inside* the main function

fvukelic: Hey @tzumainn, to run the integration and system tests, I need the application context (used in tests/base_test.py), if I put these 2 lines in the main function I don't have access to it from other files. Any suggestions on how to resolve that?

flocx_market/tests/system/test_offer.py
```diff
@@ -0,0 +1,89 @@
...
+                                 'Alice1992',
+                                 now, 'available', 'Dell380', now,
+                                 now,
+                                 {
+                                     "new attribute XYZ": "This is just a sample list of free-form attributes used for describing a server.",
+                                     "cpu_type": "Intel Xeon", "cores": 16, "ram_gb": 512,
+                                     "storage_type": "samsung SSD", "storage_size_gb": 204},
+                                 11.5)
+                offer.save_to_db()
+                res = client.get(f'/offers/{offer.marketplace_offer_id}')
```
tzumainn: There's a typo here - an extra `f`

tzumainn: Looks like this happens a couple of times in this file

flocx_market/cmd/api.py
```diff
@@ -11,19 +11,35 @@
...
-    launcher = service.ProcessLauncher(CONF, restart_method='mutate')
-    server = wsgi_service.WSGIService('flocx_market_api')
-    launcher.launch_service(server, workers=server.workers)
-    launcher.wait()
+    # launcher = service.ProcessLauncher(CONF, restart_method='mutate')
+    # server = wsgi_service.WSGIService('flocx_market_api')
+    # launcher.launch_service(server, workers=server.workers)
+    # launcher.wait()
+    orm.init_app(application)
+    application.run(port=8080, debug=True)
```
tzumainn: instead of `8080` you should have `CONF.api.port`

flocx_market/tests/system/test_offer.py
fvukelic: that is a new f string that python 3.6  and above support: https://realpython.com/python-f-strings/

tzumainn: Ah, got it - yeah, I was using the wrong version of python

flocx_market/cmd/api.py
tzumainn: Can you just write a function that returns it, instead of importing it?

fvukelic: that will probably work for now, I'll do that

tzumainn:
There are some failing tests:

* The commit message checks show that there should be some updates there; take a look at the messages for details
* The failing unit tests are because those tests can't find the config option. For now, something like https://github.com/CCI-MOC/flocx-market/pull/39/files#diff-208da164a144972de5e5585412279a27R9 should work to add those options to the test temporarily. In a future PR we really need to move this database configuration out of the API and maintain separation of layers.

tzumainn:
Awesome, thanks! This is a good foundation to build off of. Thank you for your patience!

PR #31, opened 2019-07-02
Implemented Bid API

fvukelic:
Wrote some additional tests for the Bid API and added the necessary endpoints.
- TG-66
- TG-67
- TG-68
- TG-69

### DIFF
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -2,6 +2,7 @@ from flask import Flask
 from flask_restful import Api
 from flocx_market.api.offer import Offer
 from flocx_market.api.root import Root
+from flocx_market.api.bid import Bid
 import flocx_market.conf
 
 from keystonemiddleware import auth_token
@@ -21,6 +22,10 @@ def create_app(app_name):
                      '/offer',
                      '/offer/',
                      '/offer/<string:marketplace_offer_id>')
+    api.add_resource(Bid,
+                     '/bid',
+                     '/bid/',
+                     '/bid/<string:marketplace_bid_id>')
     api.add_resource(Root, '/')
 
     if CONF.api.auth_enable:
--- a/flocx_market/api/bid.py
+++ b/flocx_market/api/bid.py
@@ -1,0 +1,47 @@
+from flask_restful import Resource
+from flocx_market.db.sqlalchemy.bid_api import BidApi
+from flask import request
+
+
+class Bid(Resource):
+
+    @classmethod
+    def get(cls, marketplace_bid_id=None):
+        if marketplace_bid_id is None:
+            return BidList.get()['bids']
+        bid = BidApi.find_by_id(marketplace_bid_id)
+        if bid:
+            return bid.as_dict()
+        return {'message': 'Bid not found'}, 404
+
+    @classmethod
+    def post(cls):
+        data = request.get_json(force=True)
+        bid = BidApi(**data)
+        bid.save_to_db()
+        return bid.as_dict(), 201
+
+    @classmethod
+    def delete(cls, marketplace_bid_id):
+        bid = BidApi.find_by_id(marketplace_bid_id)
+        if bid:
+            bid.delete_from_db()
+            return {'message': 'Bid deleted.'}
+        return {'message': 'Bid not found.'}, 404
+
+    @classmethod
+    def put(cls, marketplace_bid_id):
+        data = request.get_json(force=True)
+        bid = BidApi.find_by_id(marketplace_bid_id)
+        if bid is None:
+            return {'message': 'Bid not found.'}, 404
+
+        bid.status = data['status']
+        bid.save_to_db()
+        return bid.as_dict()
+
+
+class BidList(Resource):
+    @classmethod
+    def get(cls):
+        return {"bids": [x.as_dict() for x in BidApi.query.all()]}
--- a/flocx_market/db/sqlalchemy/bid_api.py
+++ b/flocx_market/db/sqlalchemy/bid_api.py
@@ -1,0 +1,27 @@
+from flocx_market.db.orm import orm
+from flocx_market.db.sqlalchemy.bid_model import BidModel
+
+
+class BidApi(BidModel):
+
+    def as_dict(self):
+        d = {}
+        for col in self.__table__.columns:
+            if col.name in ['start_time', 'end_time']:
+                d[col.name] = getattr(self, col.name).isoformat()
+            else:
+                d[col.name] = getattr(self, col.name)
+        return d
+
+    @classmethod
+    def find_by_id(cls, marketplace_bid_id):
+        return cls.query.filter_by(
+            marketplace_bid_id=marketplace_bid_id).first()
+
+    def save_to_db(self):
+        orm.session.add(self)
+        orm.session.commit()
+
+    def delete_from_db(self):
+        orm.session.delete(self)
+        orm.session.commit()
--- a/tox.ini
+++ b/tox.ini
@@ -29,7 +29,7 @@ commands =
         pytest --cov=flocx_market {posargs:flocx_market/tests/unit}
 
 [flake8]
-#ignore = E501,F401,E211
+#ignore = E501
 
 
 [testenv:genconfig]
--- a/flocx_market/db/sqlalchemy/bid_model.py
+++ b/flocx_market/db/sqlalchemy/bid_model.py
@@ -1,0 +1,25 @@
+import uuid
+
+from flocx_market.db.orm import orm
+import sqlalchemy_jsonfield
+
+
+class BidModel(orm.Model):
+    __tablename__ = 'bids'
+    marketplace_bid_id = orm.Column(
+        orm.String(64),
+        primary_key=True,
+        autoincrement=False,
+        default=lambda: uuid.uuid4().hex,
+    )
+    creator_bid_id = orm.Column(orm.String(64), nullable=False)
+    creator_id = orm.Column(orm.String(64), nullable=False)
+    server_quantity = orm.Column(orm.Integer, nullable=False)
+    start_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    end_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    duration = orm.Column(orm.Integer, nullable=False)
+    status = orm.Column(orm.String(15), nullable=False, default='available')
+    server_config_query = orm.Column(sqlalchemy_jsonfield.JSONField(
+                                     enforce_string=True,
+                                     enforce_unicode=False), nullable=False)
+    cost = orm.Column(orm.Float, nullable=False)
--- a/flocx_market/tests/unit/api/test_app_bid.py
+++ b/flocx_market/tests/unit/api/test_app_bid.py
@@ -1,0 +1,112 @@
+import datetime
+import json
+from unittest import mock
+
+import flocx_market.conf
+from flocx_market.db.sqlalchemy.bid_api import BidApi
+
+CONF = flocx_market.conf.CONF
+
+now = datetime.datetime.utcnow()
+
+
+test_bid_1 = BidApi(marketplace_bid_id='test_bid_1',
+                    creator_bid_id="1234",
+                    creator_id="2345",
+                    server_quantity=2,
+                    start_time=now,
+                    end_time=now,
+                    duration=16400,
+                    status="available",
+                    server_config_query={'foo': 'bar'},
+                    cost=11.5)
+
+test_bid_2 = BidApi(marketplace_bid_id='test_bid_2',
+                    creator_bid_id="2345",
+                    creator_id="3456",
+                    server_quantity=2,
+                    start_time=now,
+                    end_time=now,
+                    duration=16400,
+                    status="available",
+                    server_config_query={'foo': 'bar'},
+                    cost=11.5)
+
+
+@mock.patch('flocx_market.api.bid.BidApi.query')
+def test_get_bids(mock_query, client):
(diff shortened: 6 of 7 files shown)


tzumainn:
@fvukelic can you rebase on top of master to pull in the changes from https://github.com/CCI-MOC/flocx-market/pull/33 ? That should pull in the updates to have the rest api run from a script defined by an entrypoint. You can copy that pattern to create a script for initializing the database, similar to https://github.com/openstack/ironic/blob/master/ironic/cmd/dbsync.py

flocx_market/db/sqlalchemy/bid_api.py
```diff
@@ -0,0 +1,39 @@
...
+        self.marketplace_bid_id = str(uuid4())
+        self.creator_bid_id = creator_bid_id
+        self.creator_id = creator_id
+        self.server_quantity = server_quantity
+        self.start_time = start_time
+        self.end_time = end_time
+        self.duration = duration
+        self.status = status
+        self.server_config_query = server_config_query
+        self.cost = cost
```
larsks: Let's drop this `__init__` method here just like in #e49134845c4d754f01b1c5d5fdd860cf27fccca4. The one provided by the underlying `*Model` object is more useful.

flocx_market/tests/base_test.py
```diff
@@ -0,0 +1,36 @@
+from unittest import TestCase
```
larsks: Do we actually need this file?

flocx_market/api/app.py
```diff
@@ -1,7 +1,10 @@
 from flask import Flask
 from flask_restful import Api
 from flocx_market.api.offer import Offer
+from flocx_market.api.bid import Bid
 from flocx_market.api.root import Root
+from flask import Blueprint
+from flask import jsonify
```
tzumainn: I don't think we need these two imports

larsks:
If you rebase instead of merge, you'll get rid of that extra commit and the commit checks will pass (leaving you with the pep8 failures).

flocx_market/api/app.py
larsks: @tzumainn Those now show up in the CI tests (since we merged 7ccb4892754532014286da63c5e83b5a6bd0e064).

fvukelic: And now the tests are failing :D

fvukelic: Someone create a test_offer_api that is failing @tzumainn @larsks

tzumainn: Odd, it's not failing for me (or for the CI) - is it possible there's some sort of merge conflict?

fvukelic: @tzumainn I fixed it, now it passes

fvukelic: also I think if it is communicating with the db, it should be an integration test and not a unit test.

flocx_market/tests/unit/models/test_offer_unit.py
```diff
@@ -0,0 +1,25 @@
...
+                      "ram_gb": 512, "storage_type": "samsung SSD",
+                      "storage_size_gb": 204}, 11.5)
+        self.assertEqual(o.provider_id, "b9752cc0-9bed-4f1c-8917-12ade7a6fdbe")
+        self.assertEqual(o.creator_id, "Alice1992")
+        self.assertEqual(o.marketplace_date_created, now)
+        self.assertEqual(o.status, "available")
+        self.assertEqual(o.server_id, "Dell380")
+        self.assertEqual(o.start_time, now)
+        self.assertEqual(o.end_time, now)
+        self.assertEqual(o.cost, 11.5)
```
tzumainn: I think Lars moved this file to tests/unit/db/sqlalchemy/test_offer_api.py, so there's no need to reintroduce this here.

flocx_market/api/app.py
tzumainn: Agreed. None of the app tests are communicating with the db, are they?

fvukelic: @tzumainn in tests/unit/db/sqlalchemy/test_offer_api.py all of them are

flocx_market/tests/unit/models/test_offer_unit.py
fvukelic: That file is different, calling db method. 
This one is only testing the model creation without communicating to the db

flocx_market/tests/unit/models/test_offer_unit.py
```diff
@@ -0,0 +1,25 @@
+from flocx_market.tests.unit.unit_base_test import UnitBaseTest
```
larsks: Please take this out of the pull request.  This pull request is for implementing the Bid api, and should not have any offer related funcions.

flocx_market/tests/unit/models/test_offer_unit.py
larsks: This is  identical to https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py#L23

flocx_market/api/app.py
tzumainn: Ah, but that's not an app test (testing the REST API). That's testing the database API.

flocx_market/tests/base_test.py
fvukelic: deleted it

flocx_market/tests/unit/models/test_offer_unit.py
fvukelic: sounds good, it was a confusion on my side then on what needs to go into the unit tests @tzumainn @larsks

fvukelic: I will add a bunch of more tests that are similar to the test_offer_api for bids and contracts

flocx_market/db/sqlalchemy/bid_api.py
fvukelic: the only thing with removing this is that then the marketplace_offer_id is expected as an argument. If we keep it in there, we generate the marketplace_offer_id as a UUID during the init method. How would you suggest resolving that if we remove it? @larsks

tzumainn: Something like this: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/db/sqlalchemy/api.py#L166

fvukelic: I'm not a 100% sure if removing one method to add another one is the move? and I don't know how to do the UUID generation without it

tzumainn: Having a specific `create` method that also lets you set defaults makes a lot of sense to me. It seems odd to have the `__init__` method of BidApi be the equivalent of creating a new Bid.

UUID generation can be handled through oslo utils, as seen in the linked code. In general, whenever there's an oslo library that we can use, we should *always* be using it.

fvukelic: oh okay that makes sense @tzumainn. I just realized that Lars used a different approach here: https://github.com/CCI-MOC/flocx-market/commit/e49134845c4d754f01b1c5d5fdd860cf27fccca4

what do you think?

fvukelic: @larsks sorry for the confusion there, I completely missed that you were generating the UUID on the model itself. It makes way more sense that way

tzumainn: @fvukelic I'd personally prefer for it to get in the api, and not model. We're also going to end up doing things such as defaulting project_id (taken from the context) and the like. But I don't have a strong opinion on whether that has to be done right now.

flocx_market/api/app.py
```diff
@@ -16,6 +17,9 @@ def create_app(app_name):
     )
     app.config['PROPAGATE_EXCEPTIONS'] = CONF.flask.PROPAGATE_EXCEPTIONS
     api = Api(app)
-    api.add_resource(Offer, '/offer', '/offers/<string:marketplace_offer_id>')
+    api.add_resource(Offer, '/offer', '/offers/',
+                     '/offers/<string:marketplace_offer_id>')
+    api.add_resource(Bid, '/bid', '/bids/',
+                     '/bids/<string:marketplace_bid_id>')
```
larsks: The specification requires that we use `/bid` and `/bid/<id>`, not `/bids/<id>`.

flocx_market/api/bid.py
```diff
@@ -0,0 +1,44 @@
+from flask_restful import Resource
+from flocx_market.db.sqlalchemy.bid_api import BidApi
+from flask import request
+
+
+class Bid(Resource):
+
+    @classmethod
+    def get(cls, marketplace_bid_id):
```
larsks: This will fail for `GET /bid` because `marketplace_bid_id` is a required parameter.  Change it to a keyword parameter with a default value:

```
```

Also, this is why we write tests :). The error shows up very quickly if you test the behavior against the specification.

flocx_market/api/bid.py
```diff
@@ -0,0 +1,44 @@
...
+    @classmethod
+    def delete(cls, marketplace_bid_id):
+        bid = BidApi.find_by_id(marketplace_bid_id)
+        if bid:
+            bid.delete_from_db()
+            return {'message': 'Bid deleted.'}
+        return {'message': 'Bid not found.'}, 404
+
+    @classmethod
+    def put(cls, marketplace_bid_id):
```
larsks: This should return a 404 error if the client attempts to `PUT` against an offer that doesn't exist.

flocx_market/api/offer.py
```diff
@@ -6,7 +6,10 @@
 class Offer(Resource):
 
     @classmethod
-    def get(cls, marketplace_offer_id):
+    def get(cls, marketplace_offer_id=None):
```
larsks: As discussed in #64, let's put this in a separate PR.

flocx_market/tests/unit/common/test_service.py
```diff
@@ -39,12 +39,12 @@ def test_service_cli_valid(find_config_files):
            str(args[args.index('--api-public_endpoint') + 1]))
     assert(CONF.api.api_workers ==
            int(args[args.index('--api-api_workers') + 1]))
-    assert(CONF.api.enable_ssl_api)
+    assert CONF.api.enable_ssl_api
```
larsks: This doesn't appear to be related to the Bid API.

flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
```diff
@@ -1,20 +1,19 @@
 import datetime
```
larsks: Ditto for this; let's move non-Bid stuff to separate PRs.

setup.cfg
```diff
@@ -1,3 +1,4 @@
+
```
larsks: I don't think we need this change.

tox.ini
```diff
@@ -26,10 +26,10 @@ commands =
 
 [testenv:py37]
 commands = 
-        pytest --cov=flocx_market {posargs:flocx_market/tests/unit}
+        pytest {posargs:flocx_market/tests/unit}
```
larsks: Please remove these changes from the PR as well.  Thanks!

flocx_market/db/sqlalchemy/bid_api.py
larsks: @tzumainn I kind of want default values in the model. Let me know if you have any free time tomorrow morning and we can chat?

flocx_market/api/bid.py
larsks: (I've fixed this for offers in #65)

flocx_market/db/sqlalchemy/bid_api.py
tzumainn: @larsks Sure! I've seen some interesting things in the OpenStack code, and would not mind your opinion on them.

flocx_market/api/app.py
fvukelic: I know, I talked to @SahilTikale about changing that to follow the REST conventions. Which do you think is the right approach @larsks?

larsks: I would implement it according to the spec right now, and we can talk with Sahil when he's back if you think we should update the spec.

fvukelic: Sounds good

fvukelic:
can we merge this @larsks @tzumainn?

fvukelic:
looks like there was a new merge and there's a conflicting file now, I'll change that fast and resubmit @larsks @tzumainn, thank for the approve!

larsks:
@fvukelic just noticed this still has a `/bids` endpoint, which we had talked about removing since it's contrary to the spec document.

fvukelic:
> @fvukelic just noticed this still has a `/bids` endpoint, which we had talked about removing since it's contrary to the spec document.

hmm that is really weird, I don't have it locally, I'll remove it @larsks

larsks:
Maybe I was looking at an older version of the pr (I was looking locally)?  If so, sorry for the noise!

fvukelic:
The only problem I'm seeing is that I have each model and model api separated into separate files, but @tzumainn merged a PR with having them all in a single file, so this is causing a merge conflict. Which approach should we take here @tzumainn @larsks? When I asked 2-3 weeks ago, you said it was fine if we keep them separated, let me know if this changed @tzumainn and I'll modify my PR. Thanks!

larsks:
I don't think we need to update this PR with those changes yet. Just resolve the conflict in `flocx_market/db/sqlalchemy/offer_api.py` (by deleting if after a rebase), and I think we're good to merge.

fvukelic:
@larsks, just rebased and Travis likes it

PR #32, opened 2019-07-02
Implemented contracts API

fvukelic:
Implemented endpoints for contracts. This also covers all the basic tests for the marketplace with an Sqlite database. 

- TG-71
- TG-72
- TG-73
- TG-86

### DIFF
--- a/flocx_market/db/sqlalchemy/models.py
+++ b/flocx_market/db/sqlalchemy/models.py
@@ -7,7 +7,6 @@ from flocx_market.db.orm import orm
 
 
 class FLOCXMarketBase(models.ModelBase):
-
     metadata = None
 
     def to_dict(self):
@@ -39,9 +38,10 @@ class Bid(Base):
     duration = orm.Column(orm.Integer, nullable=False)
     status = orm.Column(orm.String(15), nullable=False, default='available')
     server_config_query = orm.Column(sqlalchemy_jsonfield.JSONField(
-                                     enforce_string=True,
-                                     enforce_unicode=False), nullable=False)
+        enforce_string=True,
+        enforce_unicode=False), nullable=False)
     cost = orm.Column(orm.Float, nullable=False)
+    contracts = orm.relationship('Contract', lazy='dynamic')
 
 
 class Offer(Base):
@@ -65,3 +65,25 @@ class Offer(Base):
         nullable=False,
     )
     cost = orm.Column(orm.Float, nullable=False)
+    contract_id = orm.Column(orm.String(64),
+                             orm.ForeignKey('contracts.contract_id'),
+                             nullable=True)
+    contract = orm.relationship('Contract')
+
+
+class Contract(Base):
+    __tablename__ = 'contracts'
+    contract_id = orm.Column(
+        orm.String(64),
+        primary_key=True,
+        autoincrement=False,
+    )
+    time_created = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    status = orm.Column(orm.String(15), nullable=False, default='available')
+    start_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    end_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
+    cost = orm.Column(orm.Float, nullable=False)
+    bid_id = orm.Column(orm.String(64),
+                        orm.ForeignKey('bids.marketplace_bid_id'))
+    bid = orm.relationship('Bid')
+    offers = orm.relationship('Offer', lazy='dynamic')
--- a/flocx_market/tests/unit/api/test_app_contract.py
+++ b/flocx_market/tests/unit/api/test_app_contract.py
@@ -1,0 +1,165 @@
+import datetime
+import json
+from unittest import mock
+
+import flocx_market.conf
+from flocx_market.db.sqlalchemy import models
+
+
+CONF = flocx_market.conf.CONF
+
+now = datetime.datetime.utcnow()
+
+
+contract_1_bid = models.Bid(marketplace_bid_id='test_bid_1',
+                            creator_bid_id="1234",
+                            creator_id="2345",
+                            server_quantity=2,
+                            start_time=now,
+                            end_time=now,
+                            duration=16400,
+                            status="available",
+                            server_config_query={'foo': 'bar'},
+                            cost=11.5)
+
+contract_2_bid = models.Bid(marketplace_bid_id='test_bid_2',
+                            creator_bid_id="2345",
+                            creator_id="3456",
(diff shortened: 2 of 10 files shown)


flocx_market/tests/system/test_offer.py
```diff
@@ -0,0 +1,89 @@
+from flocx_market.db.sqlalchemy.offer_api import OfferApi
```
larsks: This doesn't seem to be related to implementing the contract api. Maybe open an issue for adding new tests and then add it in a new pull request?

flocx_market/tests/unit/models/test_offer_unit.py
```diff
@@ -0,0 +1,21 @@
+from flocx_market.tests.unit.unit_base_test import UnitBaseTest
```
larsks: Ditto on this one.

flocx_market/tests/integration/models/test_offer_integration.py
```diff
@@ -0,0 +1,24 @@
+from flocx_market.db.sqlalchemy.offer_api import OfferApi
```
larsks: ...and this one too.

flocx_market/db/sqlalchemy/contract_api.py
```diff
@@ -0,0 +1,48 @@
...
+    def as_dict(self):
+        d = {}
+        for col in self.__table__.columns:
+            if col.name in ['time_created', 'start_time', 'end_time']:
+                d[col.name] = getattr(self, col.name).isoformat()
+            else:
+                d[col.name] = getattr(self, col.name)
+        offer_model_objects = [offer for offer in self.offers.all()]
+
+        # needed to avoid foreign key constraint issue since model and api are separated
```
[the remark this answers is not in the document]
tzumainn: I agree; this potentially makes the function very expensive database-wise.

flocx_market/db/sqlalchemy/models.py
```diff
@@ -65,3 +66,48 @@ class Offer(Base):
...
+    def contract_to_dict(self):
+        d = {}
+        for col in self.__table__.columns:
+            val = getattr(self, col.name)
+            if type(val) == datetime.datetime:
+                d[col.name] = val.isoformat()
+            else:
+                d[col.name] = val
+
+        offer_model_objects = [offer for offer in self.offers.all()]
```
tzumainn: This seems pretty dangerous; you're iterating through every single offer in the database, and if you're fetching, say, all contracts, you're doing that for every single contract.

I don't think this expansion needs to be done at all; just print the offer_id. I'd suggest removing the `contract_to_dict` function, and just using `to_dict`.

fvukelic: Hey @tzumainn, yes, sure I can make those separate commits
I call self.offers.all(), that is only iterating through all the offers related to this contract, not all in the database. Should I expand the basic to_dict method or how do you think I should approach returning of offer_ids related to this contract?

tzumainn: Whoops, you're right!

I don't think you should expand the offer at all here; the current to_dict method is sufficient for simply having the value for offer_id, correct?

fvukelic: I'm not storing the offers in the contracts because there can be many offers that match 1 bid and we don't know the number. So instead I just keep a foreign key on offers for contracts and that's why I need to pull them with self.offers. What do you think it's the best solution in that case @tzumainn?

tzumainn: Ah - I think there may be a misunderstanding here. We shouldn't need to keep track of all the offers that match a single bid; the matcher will assign one of those offers to that bid, and create a contract.

It's possible that we haven't gotten the data model quite right yet, so if we change it in the future, that's fine. Either way, I think we just want to use a generic `to_dict` function to return to users through the REST API. Don't worry about following relationships or anything yet.

fvukelic: hmm okay, I'm not keeping track of them, I am receiving a list of offers from the matcher when a contract is being made, and then I update those offers to reference the contract that is passed in. But I can't keep the offers in the contract table since there can be many offers to one contract. In the API documentation it says that the contracts should return the list of offers that it is associated with, should I just leave that out then @tzumainn?

tzumainn: I may have misunderstood; I thought you meant offers matching bids, but you mean offers matching contracts. But I thought that if a bid is looking for 5 nodes, then 5 separate contracts are created for each node - in other words, each contract specifies exactly one node?

fvukelic: yes, sorry I was unclear there. Hmm, not sure, I was following the API documentation, there the contracts can have multiple offers and 1 bid only @tzumainn

tzumainn: Hm, interesting, maybe I'm wrong! In any case, I think my general point stands: the to_dict method here should just change the database object into a dict; it shouldn't do any additional object expansion.

tzumainn:
I'm not entirely sure about the data model, but I think that's more me getting confused than anything else. I still don't think an offer should tie specifically into one contract, because it seems to me that an offer should be able to handle multiple contracts over its lifespan - but I know that's a point of contention.

In any case, I'm fine with this going in as-is, and possibly adjusting the data model later if needed. @larsks what do you think?

flocx_market/tests/unit/api/test_app_contract.py
```diff
@@ -0,0 +1,165 @@
...
+                          start_time="2016-07-16T19:20:30",
+                          end_time="2016-07-16T19:20:30",
+                          cost=0.0,
+                          bid_id='test_bid_2',
+                          offers=[contract_1_offer.marketplace_offer_id]
+                          )
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.contract_get_all')
+def test_get_bids(mock_get_all, client):
```
larsks: This should probably be `test_get_contracts` :)

tzumainn:
Ah, we have to add `contract_id` to offers in a few places:

* flocx_market/objects/offer.py - add it to the fields
* flocx_market/tests/unit/api/test_app_offer.py, flocx_market/tests/unit/objects/test_object_offer.py - add contract_id to the test offers


*

tzumainn:
Whoops, you did update objects/offer.py. It possibly has to have 'nullable=True'

fvukelic:
@tzumainn had to add it to a few places, just now saw that the tests were failing :D they passed now

tzumainn:
@fvukelic awesome, thanks! Looks good - merging now!

PR #33, opened 2019-07-02
Implement common service to start up REST API

DanNiESh:
Run cmd: flocx-market-api, then the app is start by the script cmd/start_service.py
- TG-86

### DIFF
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -1,20 +1,25 @@
-import pecan
-from pecan import make_app
-from flocx_market import model
-from flocx_market.api import config as api_config
+import os
+from flask import Flask
+from flask import Blueprint
+from flask import jsonify
+import flocx_market.conf
+CONF = flocx_market.conf.CONF
+flocx_market_api_bp = Blueprint("flocx_market_api_bp", __name__)
 
 
-def get_pecan_config():
-    filename = api_config.__file__.replace('.pyc', '.py')   # get the absolute path of the pecan config.py
-    return pecan.configuration.conf_from_file(filename)
+@flocx_market_api_bp.route('/')
+def index():
+    flocx_market_url = CONF.api.host_ip + ':' + str(CONF.api.port)
+    version = {"versions":
+               {"values": [{"status": "in progress",
+                            "updated": "2019-07-02T00:00:00Z",
+                            "media-types": [{"base": "application/json", "name": "flocx-market"}],
+                            "links": [{"href": flocx_market_url, "rel": "self"}]}]
+                }}
+    return jsonify(version)
 
 
-def setup_app():
-    config = get_pecan_config()
-    model.init_model()
-    app_conf = dict(config.app)
-    app = pecan.make_app(
-        app_conf.pop('root'),
-        logging=getattr(config, 'logging', {}),
-        **app_conf)
+def create_app(app_name):
+    app = Flask(app_name)
+    app.register_blueprint(flocx_market_api_bp)
     return app
--- a/flocx_market/api/service.py
+++ b/flocx_market/api/service.py
@@ -1,0 +1,51 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from oslo_concurrency import processutils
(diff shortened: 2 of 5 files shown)


flocx_market/api/app.py
```diff
@@ -1,20 +1,16 @@
...
 
+my_bp = Blueprint("my_bp", __name__)
 
-def get_pecan_config():
-    filename = api_config.__file__.replace('.pyc', '.py')   # get the absolute path of the pecan config.py
-    return pecan.configuration.conf_from_file(filename)
 
+@my_bp.route('/')
+def index():
+    return "Hello World"
```
tzumainn: Most OpenStack services serve something at the index; for example, if you curl your keystone endpoint, you'll see something like this:

```
[stack@localhost controllers(keystone_admin)]$ curl http://192.168.1.2:5000
{"versions": {"values": [{"status": "stable", "updated": "2019-01-22T00:00:00Z", "media-types": [{"base": "application/json", "type": "application/vnd.openstack.identity-v3+json"}], "id": "v3.12", "links": [{"href": "http://192.168.1.2:5000/v3/", "rel": "self"}]}]}}
```

I'd update this to something similar.

tzumainn:
I'd also suggest changing the commit message from "starting up the app" to "starting up the REST API". Remember, the REST API is just one application that will run from this code; there could be others.

tzumainn:
You probably also want tests for the API. Here's the pecan version: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/tests/api/base.py - you may want to look at how keystone tests its API.

tzumainn:
Oh, and the failing CI is because you need a body for your commit message.

DanNiESh:
> I'd also suggest changing the commit message from "starting up the app" to "starting up the REST API". Remember, the REST API is just one application that will run from this code; there could be others.

Sure. But the REST api part is in Filip's branch, not merged in upstream. Should I take some code from Filip?

tzumainn:
I think we need to decide which PR is dependent upon the other. My preference would be for this to get in first, and then for Filip's PR to rebase on top of this.

flocx_market/api/app.py
```diff
@@ -1,20 +1,17 @@
-import pecan
-from pecan import make_app
-from flocx_market import model
-from flocx_market.api import config as api_config
+import os
+from flask import Flask
+from flask import Blueprint
+from flask import jsonify
+my_bp = Blueprint("my_bp", __name__)
```
tzumainn: I should have caught this before, but we probably want to rename this to something a bit less casual - flocx_market_api_bp or something

flocx_market/api/app.py
```diff
@@ -1,20 +1,17 @@
...
 
 
-def get_pecan_config():
-    filename = api_config.__file__.replace('.pyc', '.py')   # get the absolute path of the pecan config.py
-    return pecan.configuration.conf_from_file(filename)
+@my_bp.route('/')
+def index():
+    version = {"versions": {"values": [{"status": "in progress", "updated": "2019-07-02T00:00:00Z", "media-types": [{"base": "application/json", "name": "flocx-marketplace"}], "links": [{"href": "http://localhost:8081", "rel": "self"}]}]}}
```
tzumainn: This is probably more readable if you put the lines of the dict on different lines. I think we're also calling it 'flocx-market' rather than 'flocx-marketplace'; and instead of hardcoding `http://localhost:8081` we should get it from oslo config.

DanNiESh: Should I modify default value for CONF.api.public_endpoint and use it to get the flocx-market url? Then I need to modify some places in config test, since the default value is None.

tzumainn: The WSGI service starts up using these parameters, right?

```
            host=CONF.api.host_ip,
            port=CONF.api.port,
```

Just use those.

DanNiESh: Thank you! All fixed.

flocx_market/api/app.py
```diff
@@ -1,20 +1,26 @@
...
-    filename = api_config.__file__.replace('.pyc', '.py')   # get the absolute path of the pecan config.py
-    return pecan.configuration.conf_from_file(filename)
+@flocx_market_api_bp.route('/')
+def index():
+    flocx_market_url = CONF.api.host_ip + ':' + str(CONF.api.port)
+    version = {
+        "versions":
+        {"values": [{"status": "in progress",
+         "updated": "2019-07-02T00:00:00Z",
+         "media-types": [{"base": "application/json", "name": "flocx-marketplace"}],
```
tzumainn: I think we should rename flocx-marketplace to just flocx-market

DanNiESh:
> Oh, and I think there's a few pep8 complaints that should probably be fixed!

Yes! I'm trying to modify the json pep8 format.

DanNiESh:
> One last minor comment; and I think you need to add Flask to requirements.txt. After that, I think this is good to merge!

requirements.txt, pep8, flocx-marketplace->flocx-market problems fixed.

tzumainn:
Looks good!

PR #36, opened 2019-07-03
Add __init__.py to flocx-market/cmd

ljmcgann:
Added an __init__.py file to the cmd directory so that when installing
the package thorugh setup.py the console_endpoint will be registered
correctly. TG-64


tzumainn:
Hi! This looks good; could you just rename the commit message into something a bit easier for an outside devleoper to understand? Something as simple as 'Add __init__.py to flocx_market/cmd' would work, I think.

PR #39, opened 2019-07-03
Add Keystone authentication to REST API

ljmcgann:
Uses the keystonemiddleware package to require any user to have a valid
token in order to query the api. Need to add tests for this still (and
could use direction in doing so as this commit isn't self contained. It
depends on the config file at /etc/flocx-market as well as deploying
an application on a port)

### DIFF
--- a/flocx_market/cmd/api.py
+++ b/flocx_market/cmd/api.py
@@ -27,3 +27,7 @@ def main():
     server = wsgi_service.WSGIService('flocx_market_api')
     launcher.launch_service(server, workers=server.workers)
     launcher.wait()
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ etc/flocx-market/flocx-market.conf.sample
 *.pyc
 __pycache__/
 .tox/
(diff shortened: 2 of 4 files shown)


flocx_market/cmd/api.py
```diff
@@ -27,3 +27,7 @@ def main():
     server = wsgi_service.WSGIService('flocx_market_api')
     launcher.launch_service(server, workers=server.workers)
     launcher.wait()
+
+
+if __name__ == '__main__':
+    main()
```
tzumainn: Use `sys.exit(main())` instead.

ljmcgann: This only exists so that I do not have to reinstall the package every single time I want to start the server. I will move this code to a module that I will just not stage for commit

tzumainn: I think this code is fine! Just do something more like: https://github.com/openstack/ironic/blob/master/ironic/cmd/api.py#L48

tzumainn:
Thinking it about, you could probably do a test where you don't set the auth_strategy to noauth, and just verify that the app rejects the request.

tzumainn:
Looks good - thanks!

PR #41, opened 2019-07-03
test flask app

DanNiESh:
test default values for wsgi service in api/service.py 
I only make a test for wsgi_server with default value. The current test may not be sufficient and needs more suggestions.
- TG-86

### DIFF
--- a/flocx_market/tests/unit/api/base.py
+++ b/flocx_market/tests/unit/api/base.py
@@ -1,0 +1,37 @@
+from flocx_market.api import app
+import flocx_market.conf
+import flask
+import unittest
+from flask import jsonify
+
+CONF = flocx_market.conf.CONF
+
+
+class BasicTests(unittest.TestCase):
+
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


tzumainn:
Hi! I think we may want to do something a little bit different. Take a look at what https://github.com/openstack/keystone/blob/master/keystone/tests/unit/core.py does in terms of setting up a test flask app to test against. I could imagine putting similar code in our tests/unit/api/base/ (other unit tests should not need this functionality), and testing against that app.

flocx_market/tests/unit/api/base.py
```diff
@@ -0,0 +1,37 @@
...
+    # executed after each test
+
+    def tearDown(self):
+        pass
+
+    def test_home_status_code(self):
+        response = self.app.get("/", follow_redirects=True)
+        self.assertEqual(response.status_code, 200)
+
+    def test_home_data(self):
```
tzumainn: I'd call this `test_root_data`

PR #63, opened 2019-07-10
Manager Service Implementation

ljmcgann:


larsks:
:+1: for being the first code commit that includes logging :)

tzumainn:
Hi! It looks like although you've specified a periodic task, you still need to add it to a thread group, as seen here:

https://github.com/openstack/mistral/blob/master/mistral/services/expiration_policy.py#L143

I haven't found great documentation on any of this beyond what's in the code - so I might suggest just doing what esi-leap (and Ironic) does when it comes to periodic tasks.

ljmcgann:
Thank you. I just got it to work. I had been looking at how nova does this all morning with no results :/

.coverage
```diff
@@ -0,0 +1 @@
...4,25,27,31,32,34,35,39,40,41,42,46,47,48,49,51,52],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/tests/unit/manager/test_manager.py":[1,2,3,5,6,9,10,11,12,13,23,16,17,19,20,26,27],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/manager/__init__.py":[1],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/manager/service.py":[1,2,3,4,5,7,10,11,12,15,16,26,36,38,17,18,19,20,21,22,23,24,40,27,28,29,30,31,32],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/cmd/__init__.py":[1,3],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/cmd/manager.py":[1,3,5,6,7,10,13,23,14,16,17,18,19],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/conf/opts.py":[],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/cmd/dbsync.py":[],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/cmd/api.py":[],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/api/service.py":[],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/api/controllers/__init__.py":[],"/home/lmcgann/Desktop/flocx-market/new_flocx_market/flocx_market/api/controllers/v1/__init__.py":[]}}
```
tzumainn: I'm guessing this was accidentally left in as part of the commit?

ljmcgann: I didn't notice this file being generated. Will update the gitignore

flocx_market/manager/service.py
```diff
@@ -0,0 +1,43 @@
...
+            context=ctx
+        )
+
+
+class Manager(periodic_task.PeriodicTasks):
+
+    @periodic_task.periodic_task(spacing=1, run_immediately=True)
+    def bar(self, context):
+        all_offers = dbapi.offer_get_all()
+        return [x.to_dict() for x in all_offers]
```
tzumainn: I think you're actually really close to making this a real piece of functionality! Here's what I'd suggest:

* Update dbapi to include a query for `offer_get_all_by_status` - or maybe specifically `offer_get_all_unexpired`, now that I think about it
* Use that query here
* check each offer's end_time against the current time, and if end_time is greater, then set the status to `expired`

flocx_market/objects/offer.py
```diff
@@ -47,3 +47,10 @@ def save(cls, data):
         cls.status = data['status']
         db.offer_update(cls.marketplace_offer_id, data)
         return cls
+
+    @classmethod
```
tzumainn: I'd split this into two methods:

* a classmethod that's called `get_all_to_be_expired`
* a method that's called `expire`

Then, update the manager service method in `manager/service.py` to iterate through the first classmethod, and then call the second method on each of its offers. This way, if there's ever, say, a REST API for an owner to expire their offer, then we have the `expire` function ready to go.

flocx_market/objects/offer.py
```diff
@@ -47,3 +47,14 @@ def save(cls, data):
...
+    def offer_update_expire(cls):
+        expired = db.offer_get_all_to_be_expired()
+        for x in expired:
+            cls.expire(x.marketplace_offer_id)
+        return len(expired)
+
+    @classmethod
```
tzumainn: This one doesn't need to be a classmethod; the above can just be:

```
for x in expired:
   x.expire()
```

flocx_market/objects/offer.py
```diff
@@ -47,3 +47,14 @@ def save(cls, data):
...
+    def update_to_expire(cls):
+        expired = db.offer_get_all_to_be_expired()
+        for x in expired:
+            x.expire()
+        return len(expired)
+
+    def expire(self):
+        db.offer_update(self.fields['marketplace_offer_id'],
```
tzumainn: This actually just needs to be `self.marketplace_offer_id`

tzumainn:
Okay, so it turns out there's a few more issues. I've illustrated the fixes here - https://github.com/tzumainn/flocx-market/commit/1cf22a540b073e38d2e5158c5a878e140bfd5ef3:

* save() is being called incorrectly. This is something that slipped by me when I did the review of the business object layer. I'll create a fix; afterwards, you should be able to just call `offer.save()`
* the database query in the database api should only be doing a database query; grabbing all the offers and then filtering is expensive
* take a look at how I updated `objects/offer.py` and `manager/service.py` to see the suggested updates to both

I'll try and get the fix for the first issue in soon!

tzumainn:
I'm good with this! Just change the commit subject and the PR subject to match, and I'll merge!

PR #65, opened 2019-07-11
Correct behavior of the Offer REST API

### LINKED ISSUE #64: Offer api does not match specification
The specification requires that the offer endpoint be `/offer`, not `/offers`, as in:

```
api = Api(app)
api.add_resource(Offer, '/offer', '/offers/<string:marketplace_offer_id>')
api.add_resource(Root, '/')
```

Additionally, requests to `/offer` should return a list of offers; currently this results in an error.

larsks:
Requests to `offer/` should return a list of available offers. Prior to
this commit, the API would return an error for requests to `/offer`.

Closes #64

### DIFF
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -16,6 +16,6 @@ def create_app(app_name):
     )
     app.config['PROPAGATE_EXCEPTIONS'] = CONF.flask.PROPAGATE_EXCEPTIONS
     api = Api(app)
-    api.add_resource(Offer, '/offer', '/offers/<string:marketplace_offer_id>')
+    api.add_resource(Offer, '/offer', '/offer/<string:marketplace_offer_id>')
     api.add_resource(Root, '/')
     return app
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -16,6 +16,9 @@ def create_app(app_name):
     )
     app.config['PROPAGATE_EXCEPTIONS'] = CONF.flask.PROPAGATE_EXCEPTIONS
     api = Api(app)
-    api.add_resource(Offer, '/offer', '/offer/<string:marketplace_offer_id>')
+    api.add_resource(Offer,
+                     '/offer',
+                     '/offer/',
+                     '/offer/<string:marketplace_offer_id>')
     api.add_resource(Root, '/')
     return app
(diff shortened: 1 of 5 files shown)


larsks:
@fvukelic would you mind taking a look at this PR? I've modified `flocx_market/api/offer.py` to respond correctly to requests for `/offer`, but I'm not certain I've done it the right way for a `flask_restful.Resource`.

Also, I think the tests that are part of this PR may be a good model for tests against the Bid and Contract APIs.

fvukelic:
@larsks I just pushed an update to the PR where I modified a few stuff to include this. We needed to add to the paths another '/offers/' part, and I did the same for bids. So now they handle this situation where the offer-id or bid-id are None. 
With just: return OfferList.get()    it will return a dictionary with {'offers': [{offers data}]} 
so I added ['offers'] after it to remove the offers word and only return the dictionary with the offers. 
Let me know if you want me to remove the 'offers' and have the api return {'offers': [{OFFERS}]}  instead of just {OFFERS}

larsks:
> Might have not been the smartest move to also push the offers change to the Bid PR -.- should I remove it?

Yeah, make that a separate PR.

fvukelic:
> > Might have not been the smartest move to also push the offers change to the Bid PR -.- should I remove it?
> 
> Yeah, make that a separate PR.

The reason I made that a PR is that a bunch of stuff was crashing since I didn't test last time the interaction with the db when we removed the __init__method from offer_api.py. So I fixed that and then I tested and fixed up the return for get as well. Submitted the PR

larsks:
@fvukelic if you're happy with the state of this pr, you can approve it, and I will merge it.

flocx_market/api/app.py
```diff
@@ -16,6 +16,6 @@ def create_app(app_name):
     )
     app.config['PROPAGATE_EXCEPTIONS'] = CONF.flask.PROPAGATE_EXCEPTIONS
     api = Api(app)
-    api.add_resource(Offer, '/offer', '/offers/<string:marketplace_offer_id>')
+    api.add_resource(Offer, '/offer', '/offer/<string:marketplace_offer_id>')
```
fvukelic: For some reason I didn't push this, this line should be like the line below:
api.add_resource(Offer, '/offer', '/offer/', '/offer/<string:marketplace_offer_id>')

larsks: That's already corrected in #64.

larsks: Thanks; I've updated the PR.

larsks:
Thanks for the review!

PR #66, opened 2019-07-11
WIP | All matching Offers

aliraza0337:
TG-105, This returns the list of matching offers to specific criteria. This pull request all fixes some of the issues with previous matcher code i.e. previously we consider "null" as an operator rather than None. Also if a user doesn't submit a valid operator in the matching criteria, we throw an exception.

### DIFF
--- a/flocx_market/matcher.py
+++ b/flocx_market/matcher.py
@@ -1,14 +1,12 @@
 import jmespath
 import re
+from flocx_market.db.sqlalchemy.offer_api import OfferApi
 
 
 def apply_operator(val1, val2, op):
 
-    if val1 is None or val2 is None or op is None:
-        return False
-
     # null operator
-    if op == 'null':
+    if op is None and val1 is not None:
         return val1
 
     neg = False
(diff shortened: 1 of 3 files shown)


flocx_market/matcher.py
```diff
@@ -1,16 +1,17 @@
...
-        return False
-
     # null operator
-    if op == 'null':
+    if op is None and val1 is not None:
         return val1
 
+    if val1 is None or val2 is None:
+        raise ValueError
+
```
larsks: I'm not sure we should raise a ValueError in this case.  Maybe someone legitimately wants to check if a list contains a null value?  `["somelist", "contains", None]`, etc.

flocx_market/matcher.py
```diff
@@ -89,12 +92,25 @@ def apply_operator(val1, val2, op):
...
         if not apply_operator(j_val, val, op):
             return False
+
     return True
+
+
+def get_all_offers_matching_specs(specs):
+
+    all_offers = OfferList.get()['offers']
```
larsks: This is probably not the right way to get the list of offers (because the `OfferList` class is explicitly for implementing the REST API).  With the current code, you probably want to use the `OfferApi` class, as in `OfferApi.query.all()`.

PR #67, opened 2019-07-11
Corrected behavior of get and post endpoints for offer API

fvukelic:
After removing the __init__ method from offer_api, we needed to change up the post method to adjust to the new changes and have it working with the database. The get method has also been adjusted to handle get request with nothing passed in as offer id.


larsks:
@fvukelic Looks good. I've rolled this into #65.

PR #70, opened 2019-07-12
Fixed up flake8 errors and create_invalid test for the offer API

fvukelic:
Fixed up flake8 errors in in tests/unit/common/test_service.py.
Fixed up test_offer_api: changed the order of
data in data creation and test_create_invalid.py

### DIFF
--- a/flocx_market/tests/unit/common/test_service.py
+++ b/flocx_market/tests/unit/common/test_service.py
@@ -39,12 +39,12 @@ def test_service_cli_valid(find_config_files):
            str(args[args.index('--api-public_endpoint') + 1]))
     assert(CONF.api.api_workers ==
            int(args[args.index('--api-api_workers') + 1]))
-    assert(CONF.api.enable_ssl_api)
+    assert(CONF.api.enable_ssl_api is True)
 
     assert(CONF.api.max_limit == 1000)
 
     assert(CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS is False)
-    assert(CONF.flask.PROPAGATE_EXCEPTIONS)
+    assert(CONF.flask.PROPAGATE_EXCEPTIONS is True)
 
     assert(CONF.host == str(args[args.index('--host') + 1]))
 
--- a/flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
+++ b/flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
@@ -8,13 +8,13 @@ from sqlalchemy.exc import IntegrityError
 
 now = datetime.datetime.utcnow()
 test_offer_data = dict(
-    marketplace_date_created=now,
     provider_id='2345',
     creator_id='3456',
+    marketplace_date_created=now,
+    status='available',
     server_id='4567',
     start_time=now,
     end_time=now,
-    status='available',
     server_config={'foo': 'bar'},
     cost=0.0,
 )
--- a/flocx_market/tests/unit/unit_base_test.py
+++ b/flocx_market/tests/unit/unit_base_test.py
@@ -1,5 +1,0 @@
-from unittest import TestCase
-
-
-class UnitBaseTest(TestCase):
-    pass


fvukelic:
@larsks really simple PR, just moved out the stuff from the Bid PR that you requested, as soon as this one is merged, I have the Bid rebased on top so I will be submitting the updated PR for that as well.

flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
```diff
@@ -29,12 +29,12 @@ def test_offer_create(app, db, session):
...
 def test_offer_create_invalid(app, db, session):
-    data = dict(test_offer_data)
-    del data['provider_id']
-    offer = OfferApi(**data)
+    del test_offer_data['provider_id']
+    offer = OfferApi(**test_offer_data)
 
     with pytest.raises(IntegrityError):
         offer.save_to_db()
+    test_offer_data['provider_id'] = '2345'
```
larsks: I'm not clear what you're doing with these changes here. The original version was intentionally not modifying `test_offer_data` because it is used by other tests...so we create a local copy of it, and use that instead, because we can modify it as much as we want.  This change doesn't introduce any functional changes in the test, unless I'm missing something.

flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
```diff
@@ -8,13 +8,13 @@
 
 now = datetime.datetime.utcnow()
 test_offer_data = dict(
-    marketplace_date_created=now,
     provider_id='2345',
     creator_id='3456',
+    marketplace_date_created=now,
+    status='available',
```
larsks: This appears to be changing the order of the attributes, but otherwise both versions of the code are the same.  What's the reason for the change here?

flocx_market/tests/unit/common/test_service.py
```diff
@@ -39,12 +39,12 @@ def test_service_cli_valid(find_config_files):
...
     assert(CONF.api.api_workers ==
            int(args[args.index('--api-api_workers') + 1]))
-    assert(CONF.api.enable_ssl_api)
+    assert CONF.api.enable_ssl_api
 
     assert(CONF.api.max_limit == 1000)
 
     assert(CONF.flask.SQLALCHEMY_TRACK_MODIFICATIONS is False)
-    assert(CONF.flask.PROPAGATE_EXCEPTIONS)
+    assert CONF.flask.PROPAGATE_EXCEPTIONS
```
larsks: There's some inconsistent use of parentheses here.  If we don't need them to extend the `assert` statement across multiple lines, then we should drop the parentheses.  In other words: I agree with your change, but it looks like we should make the same change on additional lines.

fvukelic: The only reason I remove the parenthesis for these 2 is that flake8 is failing if I leave them, but for the others, we need the parenthesis, we only need to remove them for the single expressions ones @larsks

flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
fvukelic: Yeah that's fine, I'll remove this one

flocx_market/tests/unit/db/sqlalchemy/test_offer_api.py
fvukelic: I changed the order because that is the order that it appears in the OfferModel, so it was just easier to look at, do you think we don't need that @larsks?

larsks: That's fine; I was just curious. We can keep it.

flocx_market/tests/unit/common/test_service.py
larsks: I've updated the PR to include more consistent formatting for the `assert` lines.

Issue #73, opened 2019-07-12
The Offer database API (and other database APIs) should not be within a class


tzumainn:
Right now, in https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/db/sqlalchemy/offer_api.py, we define a OfferApi class that extends OfferModel and provides operations on the OfferModel. However, the OfferApi class may not be what we want; after all, what exactly *is* an 'OfferApi' object?

What we want a place to put any and all database operations against the OfferModel - not only CRUD on a single OfferModel, but queries that return multiple OfferModels, etc. To that end, it may make more sense to remove the OfferApi class and just have individual functions that perform the database operations we want.

PR #75, opened 2019-07-12
Removed OfferApi database model

tzumainn:
The OfferApi model doesn't make sense as a way to run database operations
on an Offer. This change removes the class and puts those operations as
single functions within flocx_market/db/sqlalchemy/offer_api.py, as well
as moving the to_dict() function into the offer model. Lastly, it adds a
test for offer updates.

Addresses #73

### DIFF
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -1,36 +1,36 @@
 from oslo_utils import uuidutils
 
 from flocx_market.db.orm import orm
-from flocx_market.db.sqlalchemy.offer_model import OfferModel
+from flocx_market.db.sqlalchemy import models
 
 
-def get(marketplace_offer_id):
-    return OfferModel.query.filter_by(
+def offer_get(marketplace_offer_id):
+    return models.Offer.query.filter_by(
         marketplace_offer_id=marketplace_offer_id).first()
 
 
-def get_all():
-    return OfferModel.query.all()
+def offer_get_all():
+    return models.Offer.query.all()
 
 
-def create(values):
+def offer_create(values):
     values['marketplace_offer_id'] = uuidutils.generate_uuid()
-    offer_ref = OfferModel(**values)
+    offer_ref = models.Offer(**values)
     orm.session.add(offer_ref)
     orm.session.commit()
     return offer_ref
 
 
-def update(marketplace_offer_id, values):
+def offer_update(marketplace_offer_id, values):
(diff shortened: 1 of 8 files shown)


flocx_market/db/sqlalchemy/api.py
```diff
@@ -0,0 +1,36 @@
+from oslo_utils import uuidutils
+
+from flocx_market.db.orm import orm
+from flocx_market.db.sqlalchemy import models
+
+
+def offer_get(marketplace_offer_id):
+    return models.Offer.query.filter_by(
+        marketplace_offer_id=marketplace_offer_id).first()
```
larsks: You could also write this, which is a little shorter:

```
return models.Offer.query.get(marketplace_offer_id)
```

As far as I can tell they end up building the same SQL expression.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -0,0 +1,36 @@
+from oslo_utils import uuidutils
+
+from flocx_market.db.orm import orm
+from flocx_market.db.sqlalchemy import models
+
+
+def offer_get(marketplace_offer_id):
```
larsks: I see that Ironic puts all the database methods inside a subclass of  `ironic.db.api`, rather than using module level methods.  What do we think about that?

flocx_market/db/sqlalchemy/models.py
```diff
@@ -1,16 +1,13 @@
-import uuid
-
 from flocx_market.db.orm import orm
 import sqlalchemy_jsonfield
 
 
-class OfferModel(orm.Model):
+class Offer(orm.Model):
```
larsks: :+1:

flocx_market/db/sqlalchemy/api.py
tzumainn: Yep! I saw that, but Nova (and Mistral) do not, so I, uh, went with the simpler direction. I don't see it impacting the usage of the database api to any great degree either way. If you feel strongly about using a subclass though, I'm happy to put it in!

flocx_market/db/sqlalchemy/api.py
tzumainn: I actually wondered about this too, because OpenStack always uses the "filter_by/first" expression rather then "get". The reason is because they distinguish between an object's ID (primary key) and UUID (human-usable ID). "get" only works on the primary key.

Incidentally, the database models should probably be updated to follow this pattern as well.

PR #76, opened 2019-07-12
Add Offer business object

aliraza0337:
Hi @tzumainn , This is just initial work, I was wondering about the output of these operations etc. For now I am just looking at @fvukelic  code and deciding the output of my code based on that.

### DIFF
--- a/flocx_market/api/bid.py
+++ b/flocx_market/api/bid.py
@@ -1,6 +1,6 @@
 from flask_restful import Resource
 from flask import request
-from flocx_market.db.sqlalchemy import api as dbapi
+from flocx_market.objects import bid
 
 
 class Bid(Resource):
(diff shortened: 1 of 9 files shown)


tzumainn:
Hi! This is a reasonable start. `base.py` should be more like https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/base.py - a place for shared functionality of the base object - for example, a to_dict() method, or to accommodate the fact that all objects have created_at and updated_at fields (although that is not yet true).

There should also be a `flocx_market/objects/offer.py`, similar to https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/offer.py. This is where we have the database functions - and note use of those database functions are changing to something a bit simpler, once https://github.com/CCI-MOC/flocx-market/pull/75 merges.

With `flocx_market/objects/offer.py` in place, the next step is to update code in the REST API layer, such as https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/api/offer.py#L22. That will no longer directly call a database function; instead it'll call your business object 'create' function.

My strongest advice is to see what esi-leap does and parallel that as much as you can.

tzumainn:
Looks good!

Issue #54, opened 2019-07-10
Database code should use oslo.db


tzumainn:
Our database code should use oslo.db to take advantage of its session handling and backend support. See https://docs.openstack.org/oslo.db/latest/user/usage.html for reference.

PR #77, opened 2019-07-15
 Update sqlalchemy code to use oslo db libraries

tzumainn:
This change updates sqlalchemy code to use oslo db libraries for
models and for sessions.
    
Addresses #54


tzumainn:
The bid api merged while I was working on this, and the bid tests will fail if the bid database code is not updated to match the changes here; so I included a commit for updating the bid code as well.

flocx_market/tests/unit/conftest.py
```diff
@@ -33,14 +36,15 @@ def client(app):
 
 @pytest.fixture()
 def db(app):
-    orm.create_all()
-    yield orm
-    orm.drop_all()
+    db_api.setup_db()
+    yield db_api
+    db_api.get_session().query(db_models.Offer).delete()
```
larsks: This fixture is used by all the database tests...not just offers. Do we really need to explicitly delete objects like this? Can we just call `drop_db()` and be done with it?

flocx_market/tests/unit/conftest.py
```diff
@@ -33,14 +36,15 @@ def client(app):
...
 
 @pytest.fixture()
 def session(db):
-    session = db.session()
+    session = db.get_session()
     yield session
-    if not session.transaction.is_active:
-        session.rollback()
+    # Guess what doesn't work with oslo db sessions...
+    # session.rollback()
```
larsks: This was necessary because some of the failing tests would result in an invalid transaction, which would cause subsequent tests to fail. Maybe there's another way around it?

flocx_market/tests/unit/conftest.py
tzumainn: You'd think so, but apparently the data remains... somehow... and causes integrity errors if you try and insert the same data into the database in subsequent tests. I checked with a few other OpenStack projects to see how they dealt with the issue, and discovered that most OpenStack projects don't test the database API. The one that I found that does has this:

https://github.com/openstack/mistral/blob/master/mistral/tests/unit/base.py#L280

flocx_market/tests/unit/conftest.py
tzumainn: From what I understand of the oslo db, rollbacks are handled automatically within a session:
https://github.com/malor/oslo.db/blob/master/oslo/db/sqlalchemy/session.py

larsks:
It looks like our tests are never calling `offer_get_all`, `bid_get_all`, or `bid_update`: https://codecov.io/gh/CCI-MOC/flocx-market/pull/77/src/flocx_market/db/sqlalchemy/api.py?before=flocx_market/db/sqlalchemy/api.py

tzumainn:
> It looks like our tests are never calling `offer_get_all`, `bid_get_all`, or `bid_update`: https://codecov.io/gh/CCI-MOC/flocx-market/pull/77/src/flocx_market/db/sqlalchemy/api.py?before=flocx_market/db/sqlalchemy/api.py

I've added tests for these

flocx_market/db/sqlalchemy/api.py
```diff
@@ -1,36 +1,103 @@
...
 
 
+CONF = flocx_market.conf.CONF
+_engine_facade = None
+
+
+def get_facade():
+    global _engine_facade
+    if not _engine_facade:
+        _engine_facade = db_session.EngineFacade.from_config(CONF)
```
larsks: What's an `EngineFacade` for?

flocx_market/tests/unit/conftest.py
larsks: In that case, shouldn't this also delete Bids in the same way?

tzumainn: Whoops, you're absolutely right; added this in. I think the tests still passed because Bids have no unique constraints.

flocx_market/db/sqlalchemy/api.py
tzumainn: It's for database session handling: https://docs.openstack.org/oslo.db/latest/user/usage.html#session-handling

ljmcgann:
Or should I commit now so the others can get the fix?

flocx_market/tests/unit/conftest.py
larsks: It bothers me seeing the models called out explicitly like this (because of how we potentially need to update this fixture every time we add a new model). How about this instead?

```diff
diff --git a/flocx_market/tests/unit/conftest.py b/flocx_market/tests/unit/conftest.py
index b88d6b02..22ec96c0 100644
--- a/flocx_market/tests/unit/conftest.py
+++ b/flocx_market/tests/unit/conftest.py
@@ -38,8 +38,6 @@ def client(app):
 def db(app):
     db_api.setup_db()
     yield db_api
-    db_api.get_session().query(db_models.Offer).delete()
-    db_api.get_session().query(db_models.Bid).delete()
     db_api.drop_db()
 
 
@@ -47,5 +45,7 @@ def db(app):
 def session(db):
     session = db.get_session()
     yield session
-    # Guess what doesn't work with oslo db sessions...
-    # session.rollback()
+    session.close()
+    engine = db_api.get_facade().get_engine()
+    for table in db_models.Base.metadata.sorted_tables:
+        engine.execute(table.delete())
```

tzumainn: You're right, that works great! Thanks for the code snippet - I've updated the PR.

PR #79, opened 2019-07-15
Keystone Authentication bugfix

ljmcgann:
This commit fixes a bug introduced with the keystone authentication.The
applications served when CONF.auth_enable was set to true was causing
issues with the wsgi_service, which was expecting a Flask App object, as
opposed to a keystonemiddleware Auth_Protocol object. TG-86

### DIFF
--- a/flocx_market/tests/unit/api/test_app.py
+++ b/flocx_market/tests/unit/api/test_app.py
@@ -2,8 +2,6 @@ import json
 import flocx_market.conf
 
 CONF = flocx_market.conf.CONF
-CONF.set_override("auth_enable", False,
-                  group='api')
 
 
 def test_root_status_code(client):
--- a/flocx_market/tests/unit/conftest.py
+++ b/flocx_market/tests/unit/conftest.py
@@ -15,6 +15,9 @@ class test_app_config:
 @pytest.fixture(scope='session')
 def app():
     CONF.clear()
+    CONF.set_override("auth_enable", False,
+                      group='api')
+
     prepare_service()
     CONF.set_override('connection', 'sqlite:///:memory:',
                       group='database')
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ eventlet==0.24.1
 flake8>= 3.7.7
 Flask>=1.0.3
 flask-migrate>=2.5.2
-Flask-RESTful>=0.3.7  
+Flask-RESTful>=0.3.7
 Flask-SQLAlchemy>=2.4.0  
 itsdangerous>=1.1.0
 Jinja2>=2.10.1
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -1,8 +1,10 @@
 from flask import Flask
 from flask_restful import Api
+from flask_migrate import Migrate
 from flocx_market.api.offer import Offer
 from flocx_market.api.root import Root
 from flocx_market.api.bid import Bid
+from flocx_market.db.orm import orm
 import flocx_market.conf
 
 from keystonemiddleware import auth_token
@@ -27,6 +29,9 @@ def create_app(app_name):
                      '/bid/',
                      '/bid/<string:marketplace_bid_id>')
     api.add_resource(Root, '/')
+
+    Migrate(app, orm)
+    orm.init_app(app)
 
     if CONF.api.auth_enable:
         app = auth_token.AuthProtocol(app, dict(CONF.keystone_authtoken))
--- a/flocx_market/api/service.py
+++ b/flocx_market/api/service.py
@@ -13,9 +13,7 @@
 from oslo_concurrency import processutils
 from oslo_service import service
 from oslo_service import wsgi
-from flask_migrate import Migrate
 from flocx_market.api import app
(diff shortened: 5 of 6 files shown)


requirements.txt
```diff
@@ -4,7 +4,9 @@ alembic>=0.8.10 # MIT
 eventlet==0.24.1
 flake8>= 3.7.7
 Flask>=1.0.3
+flask-keystone==0.2
 flask-migrate>=2.5.2
+flask-oslolog==0.1
```
tzumainn: @larsks Do you know if there are policies as to which libraries we should use? Both flask-keystone and flask-oslolog are actually Rackspace projects; are there worries that they'd end up unsupported?

tzumainn:
I'm unsure as to the usage of those flask libraries; if @larsks thinks that they're okay, then they're probably fine; otherwise I think we have to look at an alternative. One alternative may be to set a global app variable, and have a function return that?

tzumainn:
It turns out there's a simpler solution: move https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/api/service.py#L30-L31 into app.py. Try that out, and let  me know what you think! (It worked for me in testing)

ljmcgann:
This works for the test cases but not for when I try to install and run the application

ljmcgann:
The problem is all the test cases are run with auth_enable set to false. So the application is never wrapped in the keystonemiddleware. Test cases are passing but the actual application is not running

tzumainn:
I tested with my proposed fix with both auth_enable set to true and to false, and everything seemed to work as expected. Does it not work for you?

We should probably add a test where auth_enable is set to true, and we make an API request and catch the 'Auth required' (or whatever) message and the appropriate return code.

ljmcgann:
My bad I was being dumb. I got it to work now. But I'm going to try to write some kind of test case before commiting again

ljmcgann:
Or should I commit now so the others can use the fix?

tzumainn:
How about updating this PR with the fix and pointing people to it, but writing the test before merging?

flocx_market/cmd/dbsync.py
```diff
@@ -11,3 +11,7 @@ def main():
     with application.app_context():
         orm.init_app(application)
         orm.create_all()
```
ljmcgann: I have moved this whole code block to app.py as well. Otherwise we run into the same issue when trying to run dbsynch. However, I'm not sure this is the best solution as then this service does nothing the api service doesnt already do

ljmcgann: Doing so still does not fix dbsynch, but it resolves the keystone side of it

tzumainn: You don't need this at all; I'd take it out, and remove the dependencies on flask-keystone and flask-oslolog. The dbsync fix is in https://github.com/CCI-MOC/flocx-market/pull/77

ljmcgann: So I will leave this commit as is and let the other pr worry about dbsynch?

tzumainn: Yeah!

flocx_market/cmd/dbsync.py
```diff
@@ -6,4 +6,4 @@
 
 def main():
     flocx_market_service.prepare_service(sys.argv)
-    db_api.setup_db()
+db_api.setup_db()
```
tzumainn: You probably don't want this change

requirements.txt
tzumainn: @ljmcgann I don't think you need these here anymore - we don't use them, correct?

flocx_market/cmd/dbsync.py
ljmcgann: That was a mistake from copy-pasting too quick

requirements.txt
ljmcgann: Little things like this are so easy to forget

flocx_market/tests/unit/api/test_app.py
```diff
@@ -33,3 +33,17 @@ def test_root_data(client):
...
+@mock.patch('flocx_market.cmd.api.service.ProcessLauncher.wait')
+def test_keystone_app(wait, launch, prepare):
+    CONF.clear()
+    CONF.set_override("auth_enable", True,
+                      group='api')
+
+    main.main()
+
+    launch.assert_called_once()
+    wait.assert_called_once()
```
tzumainn: This is definitely a good test to have. I think it should be moved to tests/unit/cmd/test_api.py, however.

flocx_market/tests/unit/conftest.py
```diff
@@ -15,6 +15,9 @@ class test_app_config:
 @pytest.fixture(scope='session')
 def app():
     CONF.clear()
+    CONF.set_override("auth_enable", False,
+                      group='api')
+
```
tzumainn: Definitely a good call here!

requirements.txt
tzumainn: Haha, yeah, I've definitely done the same

tzumainn:
Looks good - thanks!

PR #80, opened 2019-07-15
Reformatted Bid API and fixed offers

fvukelic:
- Added the ** unpacking to offers to
make it work with the mysql db. Also
modified the tests for it.
- Reformatted Bid API to meet Openstack
standards.


flocx_market/db/sqlalchemy/api.py
```diff
@@ -13,7 +13,7 @@ def offer_get_all():
     return models.Offer.query.all()
 
 
-def offer_create(values):
+def offer_create(**values):
```
tzumainn: I don't think we want to change this signature; we just want to make sure that the API passes in a dict here.

fvukelic: @tzumainn sounds good, I'll change it to that

tzumainn: I figured out how OpenStack wants us to do this, and rolled it into my PR:

https://github.com/CCI-MOC/flocx-market/pull/77/files#diff-2d0aac4490a90e219faab216070d89ebR53

and

https://github.com/CCI-MOC/flocx-market/pull/77/files#diff-958013eb9d520a9313e7fa21b2b6599bR21

Try using that code for now!

fvukelic: @tzumainn I tried it but it's not working out when I try writing to the mysql db

tzumainn: Odd, it works for me. You probably need to try on master though, with the entire PR on top.

fvukelic: I did rebase, weird... I'll try once you guys merge https://github.com/CCI-MOC/flocx-market/pull/77/

Issue #83, opened 2019-07-16
The REST API app should not run migrations for the database


tzumainn:
Database migrations should be handled through the flocx-market-dbsync script, and not the REST API.

PR #84, opened 2019-07-16
Remove database migration from REST API app

tzumainn:
The database migration can be removed from the REST API app now
that there is a separate flocx-market-dbsync script.

Addresses #83

### DIFF
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -1,6 +1,5 @@
 from flask import Flask
 from flask_restful import Api
-from flask_migrate import Migrate
 from flocx_market.api.offer import Offer
 from flocx_market.api.root import Root
 from flocx_market.api.bid import Bid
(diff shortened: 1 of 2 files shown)


flocx_market/api/app.py
```diff
@@ -30,7 +29,7 @@ def create_app(app_name):
                      '/bid/<string:marketplace_bid_id>')
     api.add_resource(Root, '/')
 
-    Migrate(app, orm)
+#    Migrate(app, orm)
```
larsks: Just remove instead of commenting?

tox.ini
```diff
@@ -21,7 +21,7 @@ commands = {posargs}
 passenv = TEST*
 commands =
         {toxinidir}/scripts/check-sorted.sh requirements.txt test-requirements.txt
-        {toxinidir}/scripts/check-conflict-markers.sh
+#        {toxinidir}/scripts/check-conflict-markers.sh
```
larsks: :-1: Is this hitting some sort of false positive?

flocx_market/api/app.py
tzumainn: Whoooops - removed.

tox.ini
tzumainn: Whoops again - I thought it was, but it wasn't. My bad - removed this change.

PR #86, opened 2019-07-16
Create bids business object

aliraza0337:
This PR implements the Business Layer for bid API. This is WIP.

### DIFF
--- a/flocx_market/api/bid.py
+++ b/flocx_market/api/bid.py
@@ -1,6 +1,6 @@
 from flask_restful import Resource
 from flask import request
-from flocx_market.db.sqlalchemy import api as dbapi
+from flocx_market.objects import bid
 
 
 class Bid(Resource):
@@ -8,40 +8,30 @@ class Bid(Resource):
     @classmethod
     def get(cls, marketplace_bid_id=None):
         if marketplace_bid_id is None:
-            return BidList.get()['bids']
-
-        bid = dbapi.bid_get(marketplace_bid_id)
-        if bid:
-            return bid.to_dict()
-        return {'message': 'Bid not found'}, 404
+            return [x.to_dict() for x in bid.Bid.get_all()]
+        b = bid.Bid.get(marketplace_bid_id)
+        if b is None:
+            return {'message': 'Bid not found'}, 404
+        else:
+            return b.to_dict()
 
     @classmethod
     def post(cls):
         data = request.get_json(force=True)
-        bid = dbapi.bid_create(data)
-        return bid.to_dict(), 201
+        return bid.Bid.create(data).to_dict(), 201
 
     @classmethod
     def delete(cls, marketplace_bid_id):
-        bid = dbapi.bid_get(marketplace_bid_id)
-        if bid:
-            dbapi.bid_destroy(marketplace_bid_id)
-            return {'message': 'Bid deleted.'}
-        return {'message': 'Bid not found.'}, 404
+        b = bid.Bid.get(marketplace_bid_id)
+        if b is None:
+            return {'message': 'Bid not found.'}, 404
+        b.destroy()
+        return {'message': 'Bid deleted.'}
 
     @classmethod
     def put(cls, marketplace_bid_id):
         data = request.get_json(force=True)
-        bid = dbapi.bid_get(marketplace_bid_id)
-        if bid is None:
+        b = bid.Bid.get(marketplace_bid_id)
+        if b is None:
             return {'message': 'Bid not found.'}, 404
-
-        bid.status = data['status']
-        dbapi.bid_update(marketplace_bid_id, bid.to_dict())
-        return bid.to_dict()
-
-
-class BidList(Resource):
-    @classmethod
-    def get(cls):
-        return {"bids": [x.to_dict() for x in dbapi.bid_get_all()]}
+        return b.save(data).to_dict()
--- a/flocx_market/objects/base.py
+++ b/flocx_market/objects/base.py
@@ -1,0 +1,32 @@
+from oslo_log import log
+from oslo_versionedobjects import base as object_base
+import datetime
+
+LOG = log.getLogger(__name__)
+
+
+class FLOCXMarketObject(object_base.VersionedObject):
+    OBJ_SERIAL_NAMESPACE = 'flocx_market_object'
+    OBJ_PROJECT_NAMESPACE = 'flocx_market'
+
+    @staticmethod
+    def _from_db_object(obj, db_obj):
+        for key in obj.fields:
+            setattr(obj, key, db_obj[key])
+            obj.obj_reset_changes()
+        return obj
+
+    @classmethod
+    def _from_db_object_list(cls, db_objs):
+        all_objs = [cls._from_db_object(cls(), db_obj) for db_obj in db_objs]
+        return all_objs
+
+    def to_dict(self):
+        ret = dict()
+        for k in self.fields:
+            val = getattr(self, k)
+            if type(val) == datetime.datetime:
+                ret[k] = val.isoformat()
+            else:
+                ret[k] = val
+        return ret
--- a/flocx_market/objects/bid.py
+++ b/flocx_market/objects/bid.py
@@ -1,0 +1,49 @@
+import flocx_market.db.sqlalchemy.api as db
+from flocx_market.objects import base
+from flocx_market.objects import fields
+
+
+class Bid(base.FLOCXMarketObject):
+
+    fields = {
+        'marketplace_bid_id': fields.StringField(),
+        'creator_bid_id': fields.StringField(),
+        'creator_id': fields.StringField(),
+        'server_quantity': fields.IntegerField(),
+        'start_time': fields.DateTimeField(nullable=True),
+        'end_time': fields.DateTimeField(nullable=True),
+        'duration': fields.IntegerField(),
+        'status': fields.StringField(),
+        'server_config_query': fields.FlexibleDictField(nullable=True),
+        'cost': fields.FloatField()
+    }
+
+    @classmethod
+    def create(cls, data):
+        b = db.bid_create(data)
+        return cls._from_db_object(cls(), b)
+
+    @classmethod
+    def get(cls, bid_id):
+        if bid_id is None:
+            return None
+        else:
+            b = db.bid_get(bid_id)
+            if b is None:
+                return None
+            else:
+                return cls._from_db_object(cls(), b)
+
+    def destroy(cls):
+        db.bid_destroy(cls.marketplace_bid_id)
+        return True
+
+    @classmethod
+    def get_all(cls):
+        all_bids = db.bid_get_all()
+        return cls._from_db_object_list(all_bids)
+
+    def save(cls, data):
+        cls.status = data['status']
+        db.bid_update(cls.marketplace_bid_id, data)
+        return cls
--- a/flocx_market/tests/unit/objects/test_object_bid.py
+++ b/flocx_market/tests/unit/objects/test_object_bid.py
@@ -1,0 +1,66 @@
+from flocx_market.objects import bid
+import datetime
+import unittest.mock as mock
+now = datetime.datetime.utcnow()
+
+
+test_bid_1 = dict(marketplace_bid_id="123",
+                  creator_bid_id="1259a51-b4d6-497d-9f75-f56c409305c8",
+                  creator_id="12a59a51",
+                  server_quantity=2,
+                  start_time=now,
+                  end_time=now,
+                  duration=16400,
+                  status="available",
+                  server_config_query={'foo': 'bar'},
+                  cost=11.2)
+
+test_bid_2 = dict(marketplace_bid_id="1232",
+                  creator_bid_id="12a59a51-b4d6-497d-9f75-f56c409305c8",
+                  creator_id="12a9a51",
+                  server_quantity=2,
+                  start_time=now,
+                  end_time=now,
+                  duration=16400,
+                  status="available",
+                  server_config_query={'foo': 'bar'},
+                  cost=11.5)
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_create')
+def test_create(bid_create):
+    bid.Bid.create(test_bid_1)
+    bid_create.assert_called_once()
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_get')
+def test_get(bid_get):
+    b = bid.Bid(**test_bid_2)
+    bid.Bid.get(b.marketplace_bid_id)
+    bid_get.assert_called_once()
+
+
+def test_get_none():
+    ret = bid.Bid.get(None)
+    assert ret is None
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_get_all')
+def test_get_all(bid_get_all):
+    bid.Bid.get_all()
+    bid_get_all.assert_called_once()
+
+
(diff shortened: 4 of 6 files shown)


flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
+import flocx_market.db.sqlalchemy.api as db
+
+
```
tzumainn: You probably want a class here; this should look similar to: https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/offer.py#L24-L37

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
...
+
+def add_bid(data):
```
tzumainn: call it `create`

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
...
+	bid = db.bid_create(data)
+	return bid.to_dict()
+
+
+def get_bid(bid_id):
```
tzumainn: This should be a `@classmethod` and just be called `get`

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
...
+		return get_all_bids()
+	else:
+		b = db.bid_get(bid_id)
+		if b is None:
+			return None 
+		else:
+			return b.to_dict()
+
+
+def del_bid(bid_id):
```
tzumainn: `destroy` seems to be standard

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
...
+	if get_bid(bid_id):
+		db.bid_destroy(bid_id)
+		return True
+	else:
+		return False
+
+def get_all_bids():
```
tzumainn: `get_all`

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,40 @@
...
+	all_bids = db.bid_get_all()
+	return [x.to_dict() for x in all_bids]
+
+
+def update_bid(bid_id, data):
```
tzumainn: and `save`

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
...
-
+import flocx_market.objects.bid as bo
 
 class Bid(Resource):
 
     @classmethod
     def get(cls, marketplace_bid_id=None):
-        if marketplace_bid_id is None:
-            return BidList.get()['bids']
+        a = bo.get_bid(marketplace_bid_id)
```
tzumainn: If you update the code to the comments I have on objects/bid.py, this becomes `bid.Bid.get(marketplace_bid_id`

flocx_market/objects/bid.py
aliraza0337: Thanks that was my next step, after the simpler one starts working, now I will try to follow the ESI Leap code :)

flocx_market/api/bid.py
```diff
@@ -1,47 +1,35 @@
 from flask_restful import Resource
 from flask import request
-from flocx_market.db.sqlalchemy import api as dbapi
+from flocx_market.objects.bid import Bid as bo
```
tzumainn: I'd remove `as bo`, and just  do things such as `Bid.get(marketplace_bid_id)` below.

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,58 @@
...
+        if bid_id is None:
+            return cls.get_all()
+        else:
+            b = db.bid_get(bid_id)
+            if b is None:
+                return None
+            else:
+                return b.to_dict()
+
+    @classmethod
```
tzumainn: This shouldn't be a class method; it should  just do something like `db.bid_destroy(self.bid_id)`

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,58 @@
...
+            return True
+        else:
+            return False
+
+    @classmethod
+    def get_all(cls):
+        all_bids = db.bid_get_all()
+        return [x.to_dict() for x in all_bids]
+
+    @classmethod
```
tzumainn: This also should not be a `classmethod`

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,58 @@
...
+        if cls.get(bid_id):
+            db.bid_destroy(bid_id)
+            return True
+        else:
+            return False
+
+    @classmethod
+    def get_all(cls):
+        all_bids = db.bid_get_all()
+        return [x.to_dict() for x in all_bids]
```
tzumainn: We don't want to return dicts here; we want to return a list of Offer business objects.

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @classmethod
+    def create(cls, data):
+        bid = db.bid_create(data)
+        return bid.to_dict()
+
+    @classmethod
+    def get(cls, bid_id):
+        if bid_id is None:
+            return cls.get_all()
```
tzumainn: I don't know that we want to return all the bids if bid_is is None; I think we just throw an error

flocx_market/objects/bid.py
```diff
@@ -0,0 +1,58 @@
...
+        else:
+            b = db.bid_get(bid_id)
+            if b is None:
+                return None
+            else:
+                return b.to_dict()
```
tzumainn: We don't want to return a dict; take a look at https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/base.py#L32-L42 to see one way of returning an Offer object.

tzumainn:
I think this is really close! Left a few more comments. You also want to update objects/__init__.py to register the new object class (similar to https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/objects/__init__.py)

flocx_market/api/bid.py
aliraza0337: The it causes some namespace issue, as my api class is also Bid, so when it tries to call get in Bid it goes into recursion. Do you think I should change the Bid class name in Objects.

flocx_market/objects/bid.py
aliraza0337: Umm, I didn't get this if I dont make this as @classmethod then I cannot call it from API.

flocx_market/api/bid.py
tzumainn: Ohhh - I'd just import flocx_market.objects.bid then, and call `bid.Bid.get()`

flocx_market/objects/bid.py
tzumainn: I think the pattern you'd expect is something like:

```
b = bid.Bid.get(bid_id)
b.destroy()
```

aliraza0337: I might have to change the get then because I return after to_dict()  there.

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
 from flask_restful import Resource
 from flask import request
-from flocx_market.db.sqlalchemy import api as dbapi
+import flocx_market.objects.bid as bid
```
tzumainn: This can just be `from flocx_market.objects import bid`

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
...
 
     @classmethod
     def delete(cls, marketplace_bid_id):
-        bid = dbapi.bid_get(marketplace_bid_id)
-        if bid:
-            dbapi.bid_destroy(marketplace_bid_id)
+        b = bid.Bid.get(marketplace_bid_id)
+        if b is None:
+            return {'message': 'Bid not found.'}, 404
+        elif b.destroy():
```
tzumainn: Is the `elif` necessary?

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
...
-        bid.status = data['status']
-        dbapi.bid_update(marketplace_bid_id, bid.to_dict())
-        return bid.to_dict()
-
-
-class BidList(Resource):
-    @classmethod
-    def get(cls):
-        return {"bids": [x.to_dict() for x in dbapi.bid_get_all()]}
+        else:
```
tzumainn: I don't think this `else` is necessary either

flocx_market/objects/base.py
```diff
@@ -0,0 +1,38 @@
...
+from oslo_versionedobjects import base as object_base
+from flocx_market.objects import fields as object_fields
+import datetime
+
+LOG = log.getLogger(__name__)
+
+
+class MarketPlaceObject(object_base.VersionedObject):
+    OBJ_SERIAL_NAMESPACE = 'markete_place_object'
+    OBJ_PROJECT_NAMESPACE = 'markete_place'
```
tzumainn: Let's call these `flocx_market_object` and `flocx_market` just for consistency

flocx_market/objects/base.py
```diff
@@ -0,0 +1,38 @@
...
+
+    fields = {
+        'created_at': object_fields.DateTimeField(nullable=True),
+        'updated_at': object_fields.DateTimeField(nullable=True),
+    }
```
tzumainn: Take this out for now - this only works if we use the TimestampMixin in the database model (which we don't right now).

flocx_market/tests/unit/api/test_app_bid.py
```diff
@@ -38,12 +38,12 @@ def test_get_bids(mock_get_all, client):
     test_result = [test_bid_1, test_bid_2]
     mock_get_all.return_value = test_result
     response = client.get("/bid", follow_redirects=True)
-    assert response.status_code == 200
-    assert len(response.json) == 2
-    assert any(x['creator_id'] == test_bid_1.creator_id
-               for x in response.json)
-    assert any(x['creator_id'] == test_bid_2.creator_id
-               for x in response.json)
+    assert response.status_code == 404
```
tzumainn: Does this test not work?

aliraza0337: This is the one that sends get request with None bid_id.

larsks: Why are we expecting a 404 error? We expect `GET /bid` to return a list of all available bids.

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
...
 from flask import request
-from flocx_market.db.sqlalchemy import api as dbapi
+import flocx_market.objects.bid as bid
 
 
 class Bid(Resource):
 
     @classmethod
     def get(cls, marketplace_bid_id=None):
-        if marketplace_bid_id is None:
```
larsks: This change breaks the API. We expect `GET /bid` to return a list of all bids. With this change, it will instead return a 404 error.

flocx_market/tests/unit/api/test_app_bid.py
tzumainn: Ah, I think there's some misunderstanding - I'll comment above.

flocx_market/api/bid.py
```diff
@@ -1,47 +1,36 @@
...
-            return BidList.get()['bids']
```
tzumainn: We still want this; only now it returns bid.Bid.get_all() (converted to dicts) explicitly.

flocx_market/api/bid.py
```diff
@@ -1,47 +1,37 @@
...
     def get(cls, marketplace_bid_id=None):
         if marketplace_bid_id is None:
-            return BidList.get()['bids']
-
-        bid = dbapi.bid_get(marketplace_bid_id)
-        if bid:
-            return bid.to_dict()
-        return {'message': 'Bid not found'}, 404
+            return [x.to_dict() for x in bid.Bid.get_all()]
+        a = bid.Bid.get(marketplace_bid_id)
```
tzumainn: minor nit; can we use the variable `b` here instead of `a`?

flocx_market/objects/base.py
```diff
@@ -0,0 +1,32 @@
+from oslo_log import log
+from oslo_versionedobjects import base as object_base
+import datetime
+
+LOG = log.getLogger(__name__)
+
+
+class MarketPlaceObject(object_base.VersionedObject):
```
tzumainn: I should have mentioned this earlier; let's call this `FLOCXMarketObject` to keep the naming consistent

flocx_market/tests/unit/objects/test_object_bid.py
```diff
@@ -0,0 +1,64 @@
...
+                  end_time=now,
+                  duration=16400,
+                  status="available",
+                  server_config_query={'foo': 'bar'},
+                  cost=11.5)
+
+
+def test_create(app, db, session):
+    ret = bid.Bid.create(test_bid_1)
+    assert getattr(ret, 'creator_id') == '12a59a51'
```
tzumainn: These tests can actually be much simpler, since we don't actually want to insert data into a database; that's for the database unit tests.

Something like:

```
import unittest.mock as mock


@mock.patch('flocx_market.db.sqlalchemy.api.bid_create')
def test_create(bid_create):
    ret = bid.Bid.create(test_bid_1)
    bid_create.assert_called_once()
```

flocx_market/tests/unit/objects/test_object_bid.py
```diff
@@ -0,0 +1,67 @@
...
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_get_all')
+def test_get_all(bid_get_all):
+    bid.Bid.get_all()
+    bid_get_all.assert_called_once()
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_destroy')
+def test_destroy(bid_destroy):
+    ret = bid.Bid.create(test_bid_1)
```
tzumainn: You don't need this anymore.

flocx_market/tests/unit/objects/test_object_bid.py
```diff
@@ -0,0 +1,67 @@
...
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_create')
+def test_create(bid_create):
+    bid.Bid.create(test_bid_1)
+    bid_create.assert_called_once()
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_create')
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_get')
+def test_get(bid_create, bid_get):
+    ret = bid.Bid.create(test_bid_1)
```
tzumainn: You don't need this (or to mock bid_create)

flocx_market/tests/unit/objects/test_object_bid.py
```diff
@@ -0,0 +1,67 @@
...
+def test_destroy(bid_destroy):
+    ret = bid.Bid.create(test_bid_1)
+    ret.marketplace_bid_id
+    ret.destroy()
+    bid_destroy.assert_called_once()
+
+
+@mock.patch('flocx_market.db.sqlalchemy.api.bid_update')
+def test_save(bid_update):
+    a = bid.Bid.create(test_bid_1)
```
tzumainn: This can be taken out.

flocx_market/tests/unit/objects/test_object_bid.py
aliraza0337: I need the marketplace_bid_id which you get assigned when you create - not sure if that  ret = bid.Bid(**test_bid_1) will also assign marketplace_bid_id.

tzumainn:
Looks good to me, and it looks like you addressed @larsks comments so I'll merge. Thanks!

PR #100, opened 2019-07-24
Updated models with project_id

ljmcgann:
Updated the object models and corresponding business layer classes to
include a project_id field tht will ultimately be used to determine
ownership of objects. TG-163


tzumainn:
It looks like we need one more change: remove `creator_id` from all the database models (and object fields, and tests, etc).

tzumainn:
Looks good!

PR #101, opened 2019-07-24
Matcher matches available Bid & Offers

aliraza0337:
Matcher puts together appropriate bids and marketplace offers to create contracts. TG-74

### DIFF
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -64,7 +64,14 @@ def offer_get_all_unexpired(context):
             models.Offer.project_id == context.project_id).all()
 
 
+def offer_get_all_by_status(status, context):
+    return get_session().query(models.Offer)\
+        .filter(models.Offer.status == status, models.Offer.project_id ==
+                context.project_id).all()
+
+
 def offer_create(values, context):
+
     values['marketplace_offer_id'] = uuidutils.generate_uuid()
     offer_ref = models.Offer()
     offer_ref.update(values)
@@ -130,6 +137,12 @@ def bid_get_all_by_project_id(context):
 def bid_get_all_unexpired(context):
     return get_session().query(models.Bid)\
         .filter(models.Bid.status != 'expired').all()
+
+
+def bid_get_all_by_status(status, context):
+    return get_session().query(models.Bid)\
+        .filter(models.Bid.status == status, models.Bid.project_id ==
+                context.project_id).all()
 
 
 def bid_create(values, context):
@@ -207,7 +220,6 @@ def contract_create(values, context):
         contract_ref = models.Contract()
         contract_ref.update(values)
         contract_ref.save(get_session())
-
         # update foreign key for offers
         for offer_id in offers:
             ocr_data = dict(contract_id=values['contract_id'],
@@ -254,8 +266,9 @@ def offer_contract_relationship_get(context,
     if offer_contract_relationship_id is None:
         return None
     return get_session().query(models.OfferContractRelationship).filter(
-        models.OfferContractRelationship.offer_contract_relationship_id
-        == offer_contract_relationship_id).one_or_none()
+        models.OfferContractRelationship.
+        offer_contract_relationship_id ==
+        offer_contract_relationship_id).one_or_none()
 
 
 def offer_contract_relationship_get_all(context, filters=None):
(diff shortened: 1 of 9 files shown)


flocx_market/db/sqlalchemy/api.py
```diff
@@ -52,6 +52,11 @@ def offer_get_all_unexpired():
         .filter(models.Offer.status != 'expired').all()
 
 
+def offer_get_all_available():
```
tzumainn: Let's have this take in a `status` argument instead of hardcoding `available`; we can call the method `offer_get_all_by_status`

flocx_market/db/sqlalchemy/api.py
```diff
@@ -85,6 +90,11 @@ def bid_get_all_unexpired():
         .filter(models.Bid.status != 'expired').all()
 
 
+def bid_get_all_available():
```
tzumainn: Same here.

flocx_market/matcher/match_engine.py
```diff
@@ -0,0 +1,15 @@
+import matcher
+from flocx_market.objects import bid
+
+
+def prepare_contract(b_, o_):
+    pass
+
+
+def match():
+    all_available_bids = bid.Bid.get_all_available()
```
tzumainn: This would become `bid.Bid.get_all_by_status('available')`

flocx_market/matcher/matcher.py
```diff
@@ -89,12 +89,34 @@ def apply_operator(val1, val2, op):
...
         if not apply_operator(j_val, val, op):
             return False
+
     return True
+
+
+def get_all_matching_offers(specs,
+                            start_time=None,
+                            end_time=None,
+                            first=False):
```
larsks: We don't seem to be using the `first` parameter anywhere in the code right now.

aliraza0337: Yes, I just kept it so that if someome wants just one matching offer then instead going through all and return a list we can return one. But I can remove it

flocx_market/db/sqlalchemy/api.py
```diff
@@ -133,6 +139,11 @@ def bid_get_all_unexpired(context):
         .filter(models.Bid.status != 'expired').all()
 
 
+def bid_get_all_by_status(status):
```
tzumainn: I might put in a `context` parameter here (and in `offer_get_all_by_status` for consistency).  This query should eventually be scoped so that it filters by `context.project_id` if the query doesn't come from an admin.

flocx_market/matcher/match_engine.py
```diff
@@ -0,0 +1,33 @@
...
+    contract_data = dict(start_time=b.start_time,
+                         time_created=datetime.datetime.utcnow(),
+                         end_time=b.end_time,
+                         cost=b.cost,
+                         project_id=b.project_id,
+                         status='available',
+                         bid_id=b.marketplace_bid_id,
+                         offers=[x.marketplace_offer_id for x in offers_used]
+                         )
+    b.status = 'busy'
```
tzumainn: I don't think we want to set the status to `busy`, right? Or if we do, we'd just set it on line 13?

tzumainn: Oh, nevermind; b is a bid! Can I suggest using more descriptive parameter names?

flocx_market/objects/offer.py
```diff
@@ -66,3 +67,41 @@ def get_all_by_project_id(cls, context):
     def expire(self, context):
         self.status = 'expired'
         self.save(context)
+
+    @classmethod
+    def get_all_by_status(cls,
```
tzumainn: I'm not a huge fan of updating `get_all_by_status` rather than simply creating a new class method. But I think that's something that can be updated in a subsequent PR, so I don't think you need to change that now.

flocx_market/matcher/match_engine.py
aliraza0337: I renamed it as bid_ as if I name it as the bid that would conflict with BO.

flocx_market/objects/offer.py
aliraza0337: Leaving this one for now then. Do you think I should write test cases for all of this in objects because right now I am testing even this in matcher?

tzumainn: Well, I guess ideally there'd be unit tests in tests/unit/objects/test_offer_object.py (or whatever it's called) that tests permutations of this. But that work would go away once it became a database query, so I'm not sure it's worthwhile. @larsks , thoughts?

aliraza0337: Cool, so I am leaving for now ... also I rebased the PR.

larsks: I think the code in `flocx_market/objects/offer.py` should have at least basic tests in `tests/unit/objects/test_object_offer.py`, yes.

aliraza0337: They are already there, the only thing that isn't there is the test for available offers within a timeslot which have no contracts.

PR #102, opened 2019-07-24
update offer table

DanNiESh:
TG-183 add provider_offer_id field in offer table.
Provider should send an offer with the provider_offer_id value in order to create the marketplace offer.


DanNiESh:
Done.

tzumainn:
Thanks!

PR #103, opened 2019-07-25
Scoping offer objects and generating contexts in all apis

ljmcgann:
Implements checks to database api to scope calls to ensure proper access
to objects. Uses oslo_context to pass a project_id to object creation,
and when querying database, checks that oslo_context project_id matches
with the object's project_id or that the context belongs to an admin.
TG-163

### DIFF
--- a/flocx_market/api/bid.py
+++ b/flocx_market/api/bid.py
@@ -1,5 +1,5 @@
 from flask_restful import Resource
-from flask import request
+from flask import request, g
 from flocx_market.objects import bid
 
 
@@ -7,9 +7,10 @@ class Bid(Resource):
 
     @classmethod
     def get(cls, marketplace_bid_id=None):
+
         if marketplace_bid_id is None:
-            return [x.to_dict() for x in bid.Bid.get_all()]
-        b = bid.Bid.get(marketplace_bid_id)
+            return [x.to_dict() for x in bid.Bid.get_all(g.context)]
+        b = bid.Bid.get(marketplace_bid_id, g.context)
         if b is None:
             return {'message': 'Bid not found'}, 404
         else:
@@ -17,12 +18,14 @@ class Bid(Resource):
 
     @classmethod
     def post(cls):
+
         data = request.get_json(force=True)
-        return bid.Bid.create(data).to_dict(), 201
+        return bid.Bid.create(data, g.context).to_dict(), 201
 
     @classmethod
     def delete(cls, marketplace_bid_id):
-        b = bid.Bid.get(marketplace_bid_id)
+
+        b = bid.Bid.get(marketplace_bid_id, g.context)
         if b is None:
             return {'message': 'Bid not found.'}, 404
         b.destroy()
@@ -30,8 +33,9 @@ class Bid(Resource):
 
     @classmethod
     def put(cls, marketplace_bid_id):
+
         data = request.get_json(force=True)
-        b = bid.Bid.get(marketplace_bid_id)
+        b = bid.Bid.get(marketplace_bid_id, g.context)
         if b is None:
             return {'message': 'Bid not found.'}, 404
         # we only allow status field to be modified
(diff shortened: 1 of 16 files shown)


flocx_market/db/sqlalchemy/api.py
```diff
@@ -38,96 +38,124 @@ def drop_db():
...
-        marketplace_offer_id=marketplace_offer_id).first()
+def offer_get(marketplace_offer_id, context):
 
+    if context.is_admin:
+        return get_session().query(models.Offer).filter_by(
+            marketplace_offer_id=marketplace_offer_id).first()
+    else:
+        return get_session().query(models.Offer).filter_by(
+            marketplace_offer_id=marketplace_offer_id,
+            project_id=context.project_id).first()
```
tzumainn: Thinking about it, I don't think Offer get/get_all needs to be scoped to project; it's legitimate for a consumer to browse all offers. I would remove this scoping from the get methods, and put it in the modify methods.

flocx_market/api/bid.py
```diff
@@ -2,36 +2,57 @@
...
 
+    @staticmethod
+    def create_context():
+        return ctx.RequestContext(
+            auth_token=request.headers['X-Auth-Token'],
+            project_id=request.headers['X-Project-Id'],
+            user_id=request.headers['X-User-Id'],
+            # is_admin=request.headers['X-Is-Admin-Project'],
+            is_admin=False,
+            overwrite=False)
```
tzumainn: Can something that's the equivalent of https://github.com/CCI-MOC/esi-leap/blob/master/esi_leap/api/app.py#L26 be done? Also, I see that this function is duplicated three times; can you put it into a common library?

flocx_market/db/sqlalchemy/api.py
ljmcgann: Should there be a get all of my offers as well as a get all offers in the db api? Or is that kind of filtering something that should happen at the ui level

tzumainn: Good question. I might suggest a 'get_all_by_project_id' function.

flocx_market/api/bid.py
ljmcgann: I am able to do ctx.RequestContext.from_environ(request.environ), but this has the opposite issue as the previous way. Every user is not admin now. Before it made every user an admin

ljmcgann: Regardless of this issue, this still seems like the intended way to do this. Will update when I finish the offer part.

flocx_market/db/sqlalchemy/api.py
ljmcgann: Should this be part of the flask api?

ljmcgann: There would need to be 2 parameters in the get function. This also goes against the api documentation sahil has provided, doesn't it?

flocx_market/api/bid.py
tzumainn: Okay, sounds good!

flocx_market/db/sqlalchemy/api.py
tzumainn: For now, I'd suggest creating the get_all_by_project_id function in the database API and the business object layer, but leaving the REST API alone. We can decide later what we want the exact behavior to be.

I'd consider the API documentation something of a "living document" : )

flocx_market/api/bid.py
```diff
@@ -2,36 +2,47 @@
 from flask import request
 from flocx_market.objects import bid
 
+import flocx_market.api.app as application
```
tzumainn: Minor issue thing that doesn't need to be fixed, but I'd personally just use 'from flocx_market.api import app` and then call `app.create_context()`.

flocx_market/api/bid.py
```diff
@@ -2,36 +2,47 @@
...
+
 
 class Bid(Resource):
 
     @classmethod
     def get(cls, marketplace_bid_id=None):
+
+        context = application.create_context()
```
tzumainn: It may be possible to have create_context automatically called before every request; see the `before_request` decorator and an example here: https://stackoverflow.com/questions/19531721/flask-g-user-and-before-request (there may be better examples).

PR #105, opened 2019-07-28
Fixed minor issues and added query by status option

fvukelic:


flocx_market/api/app.py
```diff
@@ -39,7 +39,9 @@ def create_app(app_name):
         '/offer_contract_relationship',
         '/offer_contract_relationship/',
         '/offer_contract_relationship/'
-        '<string:marketplace_offer_id>/<string:contract_id>')
+        '<string:marketplace_offer_id>/<string:contract_id>',
+        '/offer_contract_relationship/'
+        '<string:marketplace_offer_id>/<string:contract_id>/<string:status>')
```
tzumainn: Would it be possible to not create a new URL here, but to look at query parameters using request.args? Then you would just have to see if marketplace_offer_id, or contract_id, or status exists; and filter depending on whether they do.

tzumainn: Oh, and correct me if I'm misunderstanding how this URL works!

fvukelic: Hey, yes it's possible, I was just keeping the same pattern, but I can change it to that. I'll update it

DanNiESh: Hey. If the provider want to pull contracts by status, the marketplace has the contract_offer_relationship api to filter all contracts with status "unretrieved" and return only the contracts uuid. But the provider wants the contract information, like create time, price and so on. Can you do something like, when the contract_offer_relationship api has one parameter: status, then marketplace firstly queries the contract_offer_relationship table to get the contracts uuid, then queries contracts table to return these contract json data?

fvukelic: I don't think that should be handled by the marketplace. After querying the offer_contract_relationships by status, the UUIDs of the corresponding contracts will be returned which can then be used to query the contracts. I think it's cleaner that way. 
What are your thoughts @tzumainn?

fvukelic: @tzumainn, fixed up the URL thingy you asked for

tzumainn: @DanNiESh @fvukelic I think a detailed query makes sense, but probably at a different URL; for example, Ironic has a nodes/ and a nodes/detail/ endpoint (as shown here: https://docs.openstack.org/api-ref/baremetal/#list-nodes-detailed)

I'd suggest a followup PR to implement the equivalent here. Otherwise, you're turning what could be one database query into 1 + <arbitrary number> queries.

flocx_market/api/offer_contract_relationship.py
```diff
@@ -6,34 +6,43 @@
...
+        status = request.args.get('status')
+
+        if (contract_id is None) and (marketplace_offer_id is None)\
+                and (status is None):
             return [x.to_dict() for x in
                     ocr.OfferContractRelationship.get_all()]
         ocrs = ocr.OfferContractRelationship.get(marketplace_offer_id,
-                                                 contract_id)
+                                                 contract_id, status)
         if ocrs is None:
```
tzumainn: I think this is close, but not quite right; take a look at https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L1666-L1685 for what may be a cleaner way to pass these filter values.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -163,15 +163,26 @@ def contract_destroy(contract_id):
...
+            and (status is None):
+        return get_session().query(models.OfferContractRelationship).filter(
+            models.OfferContractRelationship.marketplace_offer_id
+            == marketplace_offer_id)
+
+    elif (status is not None) and (contract_id is None)\
+            and (marketplace_offer_id is None):
+        return get_session().query(models.OfferContractRelationship).filter(
+            models.OfferContractRelationship.status
+            == status)
```
tzumainn: Instead of this, you can chain filters together like this: https://github.com/openstack/ironic/blob/master/ironic/db/sqlalchemy/api.py#L302

flocx_market/objects/offer_contract_relationship.py
```diff
@@ -16,16 +16,13 @@ class OfferContractRelationship(base.FLOCXMarketObject):
     }
 
     @classmethod
-    def get(cls, marketplace_offer_id, contract_id):
-        if (contract_id is None) and (marketplace_offer_id is None):
+    def get(cls, marketplace_offer_id, contract_id, status):
```
tzumainn: Ah, I think this relationship has a uuid, right? I think we should have the `get` method just return based on the uuid; and then create a `get_all` that gets a list. And instead of taking in each possible filter parameter, do something like: https://github.com/openstack/ironic/blob/master/ironic/objects/node.py#L303

flocx_market/api/offer_contract_relationship.py
fvukelic: I'm not entirely sure which one are you referring to in this method? Since all these filters are filtering by a single element and requiring the uuid, and in our case we won't know the UUID. Can you explain how did you want that done @tzumainn?

flocx_market/db/sqlalchemy/api.py
fvukelic: I'm not entirely sure which one are you referring to in this method? Since all these filters are filtering by a single element and requiring the uuid, and in our case we won't know the UUID. Can you explain how did you want that done @tzumainn?

flocx_market/api/offer_contract_relationship.py
tzumainn: So:

* split up `get` and `get_all`
* `get` returns a single offer/contract relationship; it takes the uuid as a parameter.
* `get_all` takes in a single parameter - a dictionary of filters, as the ironic code does. That dictionary of filters is created by the REST API, and unpacked in the database api layer

flocx_market/db/sqlalchemy/api.py
tzumainn: See answer above; it requires separating out `get` from `get_all`

flocx_market/api/offer_contract_relationship.py
tzumainn: The way this will be used is:

* the provider will query for all of a certain status, and create a contract for each offer/contract relationship. It will also record the relationship's uuid.
* If any update is needed, the provider has the uuid and can use that when querying or making updates.

flocx_market/db/sqlalchemy/api.py
fvukelic: I just pushed the code with the get and get_all, let me know if that's what you were thinking about @tzumainn

flocx_market/api/offer_contract_relationship.py
```diff
@@ -6,34 +6,56 @@
 class OfferContractRelationship(Resource):
 
     @classmethod
-    def get(cls, marketplace_offer_id=None, contract_id=None):
-        if (contract_id is None) and (marketplace_offer_id is None):
+    def get(cls):
```
tzumainn: I'm not terrible familiar with flask - should this be `def get(cls, offer_contract_relationship_id)`?

flocx_market/api/app.py
```diff
@@ -37,9 +37,7 @@ def create_app(app_name):
     api.add_resource(
         OfferContractRelationship,
         '/offer_contract_relationship',
-        '/offer_contract_relationship/',
-        '/offer_contract_relationship/'
-        '<string:marketplace_offer_id>/<string:contract_id>')
+        '/offer_contract_relationship/')
```
tzumainn: Should there be an `/offer_contract_relationship/<string:offer_contract_relationship_id>` resource here? We kinda just want this to work exactly like offers or contracts or bids; the only difference is that when we "get all", we check for query variables for the filter.

flocx_market/api/offer_contract_relationship.py
```diff
@@ -6,34 +6,56 @@
...
+
+        marketplace_offer_id = request.args.get('marketplace_offer_id')
+        contract_id = request.args.get('contract_id')
+        status = request.args.get('status')
+
+        possible_filters = {
+            'marketplace_offer_id': marketplace_offer_id,
+            'contract_id': contract_id,
+            'status': status
+        }
```
tzumainn: Nice! This is exactly what I wanted.

flocx_market/api/offer_contract_relationship.py
```diff
@@ -6,34 +6,56 @@
...
+        for key, value in possible_filters.items():
+            if value is not None:
+                filters[key] = value
+
+        if len(filters) == 0:
             return [x.to_dict() for x in
                     ocr.OfferContractRelationship.get_all()]
-        ocrs = ocr.OfferContractRelationship.get(marketplace_offer_id,
-                                                 contract_id)
+        ocrs = ocr.OfferContractRelationship.get_all(**filters)
```
tzumainn: Will this return an empty list if no matches without throwing an exception? That should be allowed.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -162,26 +162,49 @@ def contract_destroy(contract_id):
...
-        return get_session().query(models.OfferContractRelationship) \
-            .filter(models.OfferContractRelationship.contract_id
-                    == contract_id and
-                    models.OfferContractRelationship.marketplace_offer_id
-                    == marketplace_offer_id).first()
-
-
-def offer_contract_relationship_get_all():
-    return get_session().query(models.OfferContractRelationship).all()
+def offer_contract_relationship_get(ocr_id=None):
```
tzumainn: We should just require ocr_id.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -162,26 +162,49 @@ def contract_destroy(contract_id):
...
+
+def offer_contract_relationship_get_all(filters=None):
+    if filters is None:
+        return get_session().query(models.OfferContractRelationship).all()
+    else:
+        query_params = {}
+        for field in ['marketplace_offer_id', 'contract_id', 'status']:
+            if field in filters:
+                query_params[field] = filters[field]
+
```
tzumainn: Instead of the lines below, can't you just chain the query filters as in https://github.com/openstack/ironic/blob/master/ironic/db/sqlalchemy/api.py#L307-L308 ?

flocx_market/api/offer_contract_relationship.py
fvukelic: since we are grabbing it from request.args.get(), then we don't need it in the method parameters @tzumainn

flocx_market/api/app.py
fvukelic: yep, responded in previous comment

flocx_market/api/offer_contract_relationship.py
fvukelic: It should, I'll write a test for it

flocx_market/api/offer_contract_relationship.py
tzumainn: Is there a reason *not* to make this parallel to the way the offer/bid/contract code is written though?

flocx_market/db/sqlalchemy/api.py
fvukelic: it worked!

flocx_market/db/sqlalchemy/api.py
```diff
@@ -249,26 +249,33 @@ def contract_destroy(contract_id, context):
...
-    return get_session().query(models.OfferContractRelationship).all()
+def offer_contract_relationship_get_all(context, filters=None):
+    if filters is None:
+        return get_session().query(models.OfferContractRelationship).all()
+    else:
+        query = []
+        for field in ['marketplace_offer_id', 'contract_id', 'status']:
+            if field in filters:
+                query = get_session().query(models.OfferContractRelationship)\
+                    .filter_by(**{field: filters[field]})
```
tzumainn: Try something like this instead:

```
    query = get_session().query(models.OfferContractRelationship)
    for field in ['marketplace_offer_id', 'contract_id', 'status']:
            if field in filters:
                query = query.filter_by(**{field: filters[field]})
    return query.all()
```

Let me know if that works!

tzumainn: Oh, and theoretically this should eliminate the need for lines 271-277 below.

fvukelic: it works for the single argument queries, but the multiple arg queries are failing with it

tzumainn: Can I see the code you're using for the multiple arg queries that are failing? It worked for me with offers, so I'm confused why it'd be failing here.

flocx_market/api/offer_contract_relationship.py
```diff
@@ -6,41 +6,59 @@
...
+
+        if len(filters) == 0:
             return [x.to_dict() for x in
                     ocr.OfferContractRelationship.get_all(g.context)]
-        ocrs = ocr.OfferContractRelationship.get(
-            g.context,
-            contract_id=contract_id,
-            marketplace_offer_id=marketplace_offer_id)
 
+        ocrs = ocr.OfferContractRelationship.get_all(g.context, **filters)
```
tzumainn: I don't think we need to separate the cases where there are filters versus when there are no filters; if the latter case, passing in an empty filter dict should still just work, correct?

flocx_market/objects/offer_contract_relationship.py
```diff
@@ -16,39 +16,35 @@ class OfferContractRelationship(base.FLOCXMarketObject):
...
-            context=context)
+            context,
+            self.offer_contract_relationship_id)
         return True
 
     @classmethod
-    def get_all(cls, context):
-        all_ocrs = db.offer_contract_relationship_get_all(context)
-        return cls._from_db_object_list(all_ocrs)
+    def get_all(cls, context, filters=None):
```
tzumainn: If `filters` is a dict, the default should probably be `{}`

flocx_market/db/sqlalchemy/api.py
fvukelic: the tests are failing @tzumainn

tzumainn: Right - what code are you using when the test is failing? Is it just this code, in which case I'd suggest using my modified version, or something else?

fvukelic: yeah just this code, I run tox and it won't pass

flocx_market/api/offer_contract_relationship.py
fvukelic: yep, that's correct, removed

flocx_market/db/sqlalchemy/api.py
fvukelic: In the, /db/sqlalchemy/test_api.py the ocr tests are failing

tzumainn: Can you try the version of the code similar to what I pasted above? You set `query` *outside* of the `for` loop, and then add filters one-by-one.

fvukelic: `def offer_contract_relationship_get_all(context, filters={}):
    if len(filters) == 0:
        return get_session().query(models.OfferContractRelationship).all()
    else:
        query = get_session().query(models.OfferContractRelationship)
        for field in ['marketplace_offer_id', 'contract_id', 'status']:
            if field in filters:
                query = query.filter_by(**{field: filters[field]})
        return query.all()`

fvukelic: yeah that's what I did, I just replaced the bottom part of the code with what you pasted

fvukelic: Screen Shot 2019-08-01 at 11 11 52 AM

tzumainn: okay - what tests are failing at that point?

fvukelic: In the, /db/sqlalchemy/test_api.py the ocr tests are failing.
The problem is that when we are querying by multiple parameters and not just the uuid, the query is not getting chained on and is only returning an empty list

flocx_market/objects/offer_contract_relationship.py
larsks: @tzumainn you should never set the default for a keyword parameter to `{}` (or to `[]`).

larsks: (https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)

tzumainn: Oh, interesting  - today I have learned something!

fvukelic: okay Rebase and pushed the updated version

flocx_market/db/sqlalchemy/api.py
tzumainn: Okay, I took a look, and the tests worked for me with my suggested changes as soon as I updated a few things. Instead of:

```
def test_offer_contract_relationship_create_valid(app, db, session):
    contract_data, offer_test_id = create_test_contract_data_for_ocr()
    contract = api.contract_create(contract_data, admin_context)
    filters = {
	'marketplace_offer_id': offer_test_id,
        'contract_id': contract.contract_id
    }
    ocr = api.offer_contract_relationship_get_all(
        context=admin_context,
        filters=filters,
    )
    assert contract.contract_id == ocr.contract_id
```

I changed the last line to

```
   assert contract.contract_id == ocr[0].contract_id
```

fvukelic: that is so weird ... I changed this line `ocr = api.offer_contract_relationship_get_all(
        context=admin_context,
        filters=filters,
    )`

by adding to the end [0] and I was getting an error because it was an empty list...
Which is basically the same thing. I'm going to try your version now

fvukelic: and of course, it works now :D love when stuff like this happens. Pushed the updates @tzumainn

flocx_market/conf/api.py
```diff
@@ -13,7 +13,7 @@
     cfg.BoolOpt('enable_ssl_api',
                 default=False),
     cfg.BoolOpt('auth_enable',
-                default=True)
+                default=False)
```
tzumainn: We definitely don't want to change this default.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -41,7 +41,7 @@ def drop_db():
 def offer_get(marketplace_offer_id, context):
 
     return get_session().query(models.Offer).filter_by(
-            marketplace_offer_id=marketplace_offer_id).one_or_none()
+            marketplace_offer_id=marketplace_offer_id).first()
```
tzumainn: Ah, we're using `one_or_none()` now, after a change Lars made.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -249,26 +249,24 @@ def contract_destroy(contract_id, context):
...
+        models.OfferContractRelationship.offer_contract_relationship_id
+        == offer_contract_relationship_id).first()
 
 
-def offer_contract_relationship_get_all(context):
-    return get_session().query(models.OfferContractRelationship).all()
+def offer_contract_relationship_get_all(context, filters=None):
+    if filters is None:
+        return get_session().query(models.OfferContractRelationship).all()
+    else:
```
tzumainn: I don't think we need an `if` statement at all, just the contents of the `else`

flocx_market/objects/offer_contract_relationship.py
```diff
@@ -16,39 +16,35 @@ class OfferContractRelationship(base.FLOCXMarketObject):
...
+
+        o = db.offer_contract_relationship_get_all(context, filters)
+        if o is None:
+            return None
```
tzumainn: Do we need this? `o` would be an empty list, which would be handled fine below, right?

flocx_market/objects/offer_contract_relationship.py
```diff
@@ -16,39 +16,35 @@ class OfferContractRelationship(base.FLOCXMarketObject):
...
+
+        if filters is None:
+            return cls._from_db_object_list(o)
+        else:
+            return cls._from_db_object(cls(), o)
```
tzumainn: I don't think we want this line either; `get_all` should always return a list.

flocx_market/conf/api.py
fvukelic: got it

flocx_market/db/sqlalchemy/api.py
fvukelic: sounds good

flocx_market/api/app.py
```diff
@@ -43,15 +43,16 @@ def create_app(app_name):
...
+        '/offer_contract_relationship'
+        '/<string:offer_contract_relationship_id>')
     api.add_resource(Root, '/')
 
     orm.init_app(app)
 
     @app.before_request
     def before_request():
         g.context = ctx.RequestContext.from_environ(request.environ)
+        # g.context.is_admin = True
```
tzumainn: Ah, probably want to get rid of this : )

fvukelic: :D the sign of curl testing :D

PR #108, opened 2019-07-29
Scoping bids, contracts, and offer_contract obj

ljmcgann:
Uses oslo_context to scope queries to the bids,
contracts, and offer_contract relationship objects. Updated and added
new tests to accomodate these changes. TG-163

### DIFF
--- a/flocx_market/api/offer_contract_relationship.py
+++ b/flocx_market/api/offer_contract_relationship.py
@@ -1,5 +1,5 @@
 from flask_restful import Resource
-from flask import request
+from flask import request, g
 from flocx_market.objects import offer_contract_relationship as ocr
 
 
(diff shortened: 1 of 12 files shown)


flocx_market/api/offer_contract_relationship.py
```diff
@@ -9,9 +9,11 @@ class OfferContractRelationship(Resource):
...
         if (contract_id is None) and (marketplace_offer_id is None):
             return [x.to_dict() for x in
-                    ocr.OfferContractRelationship.get_all()]
-        ocrs = ocr.OfferContractRelationship.get(marketplace_offer_id,
-                                                 contract_id)
+                    ocr.OfferContractRelationship.get_all(g.context)]
+        ocrs = ocr.OfferContractRelationship.get(
+            marketplace_offer_id=marketplace_offer_id,
+            contract_id=contract_id,
+            context=g.context)
```
tzumainn: The context should not be optional. I'd put it as the first argument, like:

```
        ocrs = ocr.OfferContractRelationship.get(g.context,
            marketplace_offer_id=marketplace_offer_id,
            contract_id=contract_id)
```

ljmcgann: The context is not optional? I was just passing in the arguments out of order, as some of the calls to the offercontractrelationsip object were out of order. But in the method signature it is not optional. Or am I misunderstanding something

ljmcgann: the signature in the api is
def offer_contract_relationship_get(context,
                                    marketplace_offer_id=None,
                                    contract_id=None):

flocx_market/objects/offer_contract_relationship.py
```diff
@@ -16,38 +16,45 @@ class OfferContractRelationship(base.FLOCXMarketObject):
     }
 
     @classmethod
-    def get(cls, marketplace_offer_id, contract_id):
+    def get(cls, marketplace_offer_id, contract_id, context):
```
tzumainn: I'd put `context` as the first argument for consistency.

PR #109, opened 2019-07-30
Error Raising in dbapi

ljmcgann:
Creates a set of exceptions and modifies the dbapi to raise these errors
when appropriate. TG-163

### DIFF
--- a/flocx_market/common/exception.py
+++ b/flocx_market/common/exception.py
@@ -1,0 +1,29 @@
+class MarketplaceException(Exception):
+    msg_fmt = "An unknown exception occurred."
+    code = 500
+
+    def __init__(self, message=None, **kwargs):
+        if message is None:
+            try:
+                message = self.msg_fmt.format(**kwargs)
+            except KeyError:
+                message = self.msg_fmt
+
+        self.message = message
+        super(MarketplaceException, self).__init__(message)
+
+
+class ResourceNotFound(MarketplaceException):
+    code = 404
+    msg_fmt = "{resource_type} {resource_uuid} not found."
+
+
+class ResourceNoPermission(MarketplaceException):
+    code = 403
+    msg_fmt = "You do not have permissions on {resource_type} {resource_uuid}."
+
+
+class RequiresAdmin(MarketplaceException):
+    code = 403
+    msg_fmt = ("You must be an admin to perform this"
+               "action on type {resource_type}.")
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -3,6 +3,7 @@ from oslo_utils import uuidutils
 
 import flocx_market.conf
 from flocx_market.db.sqlalchemy import models
+from flocx_market.common import exception
 
 CONF = flocx_market.conf.CONF
 _engine_facade = None
@@ -40,8 +41,14 @@ def drop_db():
 
 def offer_get(marketplace_offer_id, context):
 
-    return get_session().query(models.Offer).filter_by(
+    offer_ref = get_session().query(models.Offer).filter_by(
             marketplace_offer_id=marketplace_offer_id).one_or_none()
+
+    if offer_ref:
+        return offer_ref
+    else:
+        raise exception.ResourceNotFound(resource_type="Offer",
+                                         resource_uuid=marketplace_offer_id)
 
 
 def offer_get_all(context):
@@ -102,16 +109,20 @@ def offer_update(marketplace_offer_id, values, context):
                     marketplace_offer_id=marketplace_offer_id).one_or_none()
     else:
         offer_ref = get_session().query(models.Offer).filter_by(
-            marketplace_offer_id=marketplace_offer_id,
-            project_id=context.project_id).one_or_none()
+            marketplace_offer_id=marketplace_offer_id).one_or_none()
 
     if offer_ref:
+        if offer_ref.project_id != context.project_id and not context.is_admin:
+            raise exception.ResourceNoPermission(
+                                            resource_type="Offer",
+                                            resource_uuid=marketplace_offer_id)
         values.pop('marketplace_offer_id', None)
         offer_ref.update(values)
         offer_ref.save(get_session())
         return offer_ref
     else:
-        return None
+        raise exception.ResourceNotFound(resource_type="Offer",
+                                         resource_uuid=marketplace_offer_id)
 
 
 def offer_destroy(marketplace_offer_id, context):
@@ -120,10 +131,13 @@ def offer_destroy(marketplace_offer_id, context):
                     marketplace_offer_id=marketplace_offer_id).one_or_none()
     else:
         offer_ref = get_session().query(models.Offer).filter_by(
-            marketplace_offer_id=marketplace_offer_id,
-            project_id=context.project_id).one_or_none()
+            marketplace_offer_id=marketplace_offer_id).one_or_none()
 
     if offer_ref:
+        if offer_ref.project_id != context.project_id and not context.is_admin:
+            raise exception.ResourceNoPermission(
+                                            resource_type="Offer",
+                                            resource_uuid=marketplace_offer_id)
 
         if context.is_admin:
             get_session().query(models.Offer).filter_by(
@@ -133,12 +147,20 @@ def offer_destroy(marketplace_offer_id, context):
                 marketplace_offer_id=marketplace_offer_id,
                 project_id=context.project_id).delete()
     else:
-        return None
+        raise exception.ResourceNotFound(resource_type="Offer",
+                                         resource_uuid=marketplace_offer_id)
 
 
 def bid_get(marketplace_bid_id, context):
-    return get_session().query(models.Bid).filter_by(
+
+    bid_ref = get_session().query(models.Bid).filter_by(
         marketplace_bid_id=marketplace_bid_id).one_or_none()
+
+    if bid_ref:
+        return bid_ref
+    else:
+        raise exception.ResourceNotFound(resource_type="Bid",
+                                         resource_uuid=marketplace_bid_id)
 
 
 def bid_get_all(context):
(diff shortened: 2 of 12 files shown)


flocx_market/db/sqlalchemy/api.py
```diff
@@ -39,19 +40,34 @@ def drop_db():
...
 
 def offer_get_all_by_project_id(context):
-    return get_session().query(models.Offer).filter_by(
+    offer_refs = get_session().query(models.Offer).filter_by(
         project_id=context.project_id).all()
 
+    if offer_refs:
+        return offer_refs
+    else:
+        return None
```
tzumainn: I think you could just return offer_refs, which would return an empty list in the case that it didn't match anything, right?

flocx_market/common/exception.py
```diff
@@ -0,0 +1,34 @@
...
+        self.message = message
+        super(MarketplaceException, self).__init__(message)
+
+
+class ResourceNotFound(MarketplaceException):
+    msg_fmt = "{resource_type} {resource_uuid} not found."
+
+
+class ResourceQueryNotFound(MarketplaceException):
+    msg_fmt = "Found no {resource_type}s"
```
tzumainn: This exception probably isn't necessary; querying for all of an object and getting nothing in return is legitimate.

flocx_market/db/sqlalchemy/api.py
```diff
@@ -89,7 +105,7 @@ def offer_update(marketplace_offer_id, values, context):
         offer_ref.save(get_session())
         return offer_ref
     else:
-        return None
+        raise exception.ResourceQueryNotFound(resource_type="Offer")
```
tzumainn: We'd want 'ResouceNotFound', right? The exception that prints out the uuid?

flocx_market/db/sqlalchemy/api.py
```diff
@@ -111,27 +127,47 @@ def offer_destroy(marketplace_offer_id, context):
                 marketplace_offer_id=marketplace_offer_id,
                 project_id=context.project_id).delete()
     else:
-        return None
+        raise exception.ResourceQueryNotFound(resource_type="Offer")
```
tzumainn: Same here.

flocx_market/common/exception.py
```diff
@@ -0,0 +1,34 @@
...
+        if 'code' not in self.kwargs:
+            try:
+                self.kwargs['code'] = self.code
+            except AttributeError:
+                pass
+
+        if not message:
+            try:
+                message = self.msg_fmt.format(**kwargs)
+            except Exception as e:
```
larsks: I would prefer to see us catching *specific* exceptions here. It's probably just going to be `KeyError`, right?

flocx_market/common/exception.py
```diff
@@ -0,0 +1,34 @@
+class MarketplaceException(Exception):
+    msg_fmt = "An unknown exception occurred."
+    code = 500
+    safe = False
+
+    def __init__(self, message=None, **kwargs):
+        self.kwargs = kwargs
```
larsks: We don't appear to be doing anything with `self.kwargs` after this (we're using `kwargs` on line 16).

flocx_market/common/exception.py
```diff
@@ -0,0 +1,34 @@
...
+        if 'code' not in self.kwargs:
```
larsks: I think you could replace this entire `if` block with:

```
kwargs.setdefault('code', self.code)
```

flocx_market/common/exception.py
```diff
@@ -0,0 +1,34 @@
...
+            try:
+                self.kwargs['code'] = self.code
+            except AttributeError:
+                pass
+
+        if not message:
+            try:
+                message = self.msg_fmt.format(**kwargs)
+            except Exception as e:
+                print(e)
```
larsks: We shouldn't have any `print` statements in our code. If you need to log a message, get a logging object and use that.

flocx_market/tests/unit/db/sqlalchemy/test_api.py
```diff
@@ -89,14 +89,31 @@
...
+def test_offer_get_all_found(app, db, session):
     api.offer_create(test_offer_data, scoped_context)
     api.offer_create(test_offer_data_2, scoped_context)
     api.offer_create(test_offer_data_3, scoped_context_2)
 
     assert len(api.offer_get_all(scoped_context)) == 3
 
 
+def test_offer_get_all_none_found(app, db, session):
+    with pytest.raises(Exception) as excinfo:
```
larsks: You should be testing for a *specific* exception here, not a generic `Exception`.  Ditto for all the other tests using `pytest.raises`.

flocx_market/common/exception.py
```diff
@@ -0,0 +1,24 @@
...
+                message = self.msg_fmt.format(**kwargs)
+            except KeyError:
+                message = self.msg_fmt
+
+        self.message = message
+        super(MarketplaceException, self).__init__(message)
+
+
+class ResourceNotFound(MarketplaceException):
+    msg_fmt = "{resource_type} {resource_uuid} not found."
```
larsks: Should this also set `code = 404`?

flocx_market/common/exception.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ResourceNoPermission(MarketplaceException):
+    msg_fmt = "You do not have permissions on {resource_type} {resource_uuid}."
```
larsks: ...and should this set `code = 403`?

flocx_market/db/sqlalchemy/api.py
```diff
@@ -89,7 +95,7 @@ def offer_update(marketplace_offer_id, values, context):
         offer_ref.save(get_session())
         return offer_ref
     else:
-        return None
+        raise exception.ResourceNotFound(resource_type="Offer")
```
larsks: Is this going to work?  The `ResourceNotFound` class has `msg_fmt = "{resource_type} {resource_uuid} not found."`, so this will just hit the `KeyError` in [line 12](https://github.com/CCI-MOC/flocx-market/pull/109/files#diff-a108e762755a42368a75f9314b728caeR12).

ljmcgann: It actually does not break for reasons I don't understand. Instead it returns
{resource_type} {resource_uuid} not found.
Strangely, resource_type does not get parsed, even though I did pass in that argument. I am trying to find a more elegant solution. This is initially why I had a third error class

ljmcgann: Passing in kwargs instead of **kwargs causes a Keyerror however

larsks: It returns `{resource_type} {resource_uuid} not found` because it generates ` KeyError`, since the format string expects both `resource_type` and `resource_uuid` but you're only providing `resource_type`.  Since there is no `resource_uuid`, you get a `KeyError`, and it defaults to using the raw `msg_format` instead.

ljmcgann: I figured it was some kind of error handling on behalf of format. Though why **kwargs works but not kwargs is strange. Has to do with error handling on unpacked items within the format function. Regardless, I am going to just make a third error class. This error is actually not even needed here. Mainn said get_all should just return an empty list when the queries find nothing

ljmcgann: But in the contract code this third class will be needed

ljmcgann:
Does it really make sense to return a "ResourceNotFound" error when doing a get? Currently the different get_all functions return an empty list if the specific query finds nothing
But Filip's new get_alls will accept a query filter. Combining our code will give us a get that returns an error when it does not find an offer by id, but get_all(marketplace_id==value) will return an empty list. These two functions can effectively do the same thing but with inconsistent return types

ljmcgann:
Perhaps the only error raising should be the no permissions one, and resources not being found should just return None

tzumainn:
I do think we should throw `ResourceNotFound` rather than None; if someone is specifically querying for an object based on a UUID, or trying to update it or delete it - then there's some issue going on somewhere if that object can't be found.

This looks generally good! Just needs a rebase based on lars's update.

ljmcgann:
This code should be merged after Filip merges his changes to the filters, as I will have to update my code anyways and merging will likely break all of his code again

flocx_market/common/exception.py
```diff
@@ -0,0 +1,32 @@
...
+    code = 404
+    msg_fmt = "{resource_type} {resource_uuid} not found."
+
+
+class ResourceNoPermission(MarketplaceException):
+    code = 403
+    msg_fmt = "You do not have permissions on {resource_type} {resource_uuid}."
+
+
+class RequiresAdministrativePriviledges(MarketplaceException):
```
larsks: "privileges". Or avoid the spelling issue and just call it `RequiresAdmin`.

flocx_market/common/exception.py
```diff
@@ -0,0 +1,32 @@
+class MarketplaceException(Exception):
+    msg_fmt = "An unknown exception occurred."
+    code = 500
+    safe = False
```
larsks: What is `safe` for?

flocx_market/common/exception.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, message=None, code=None, **kwargs):
```
larsks: There's a problem with the current design.  If I have code like this:

```
try:
  raise MarketplaceException(code=400)
except MarketplaceException as err:
  print('Yikes, operation failed with error code {code}'.format(err.code))
```

I will get:

```
Yikes, operation failed with error code 500.
```

There are a couple ways of dealing with this. The simplest is just to drop the `code` keyword argument: if we need a custom response code, we create a new child exception.

ljmcgann: So if I drop the code arg then we just will have a custom child class for this. Which we do anyways for all of the errors currently being raised

flocx_market/db/sqlalchemy/api.py
```diff
@@ -73,6 +80,7 @@ def offer_get_all_by_status(status, context):
 def offer_create(values, context):
 
     values['marketplace_offer_id'] = uuidutils.generate_uuid()
+    values['project_id'] = context.project_id
```
tzumainn: It turns out we don't want this; a marketplace offer is created by the provider, and we just want the provider to pass in the appropriate project_id from its own offer.

ljmcgann: But we should keep the context right? Just also take in the extra parameter?

ljmcgann: And when you say the provider should pass it in, are you saying the provider should pass it into the rest api?

ljmcgann: So I can just remove this line and the fix it good?

ljmcgann: But this line does not need to be removed from the bids as the bids are made directly with the marketplace, as opposed to going through a provider. Right?

tzumainn: Yep, just remove this line and you're good. And you're exactly correct about the bids!

ljmcgann:
Is there anything holding up this pr at this point. Mainn has approved it but it says merging is blocked for requested changed (that I addressed, I believe)

PR #110, opened 2019-07-30
Updated db models to use TimestampMixin and standardize on project_id

tzumainn:
This change uses the TimestampMixin to standardize on created_at and updated_at
for all our objects; note that these columns are automatically set and updated
by oslo.db. It also removes creator-type columns, as those are superceded by
project_id

TG-183

### DIFF
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -66,7 +66,6 @@ def offer_get_all_unexpired(context):
 
 def offer_create(values, context):
     values['marketplace_offer_id'] = uuidutils.generate_uuid()
-    values['project_id'] = context.project_id
     offer_ref = models.Offer()
     offer_ref.update(values)
     offer_ref.save(get_session())
--- a/flocx_market/db/sqlalchemy/models.py
+++ b/flocx_market/db/sqlalchemy/models.py
@@ -6,7 +6,7 @@ import datetime
 from flocx_market.db.orm import orm
 
 
-class FLOCXMarketBase(models.ModelBase):
+class FLOCXMarketBase(models.TimestampMixin, models.ModelBase):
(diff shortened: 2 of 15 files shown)


flocx_market/db/sqlalchemy/api.py
```diff
@@ -66,7 +66,8 @@ def offer_get_all_unexpired(context):
 
 def offer_create(values, context):
     values['marketplace_offer_id'] = uuidutils.generate_uuid()
-    values['project_id'] = context.project_id
+    if 'project_id' not in values:
+        values['project_id'] = context.project_id
```
larsks: Should we throw an error (or at least log a warning) if `values` has no `project_id` key?

tzumainn: Oslo.db. will throw the error for you, as the project_id can't be null.

Incidentally this is a bit of a compromise; many services which expect objects only to be created by users through some UI will always use context.project_id here. However offers are actually always created through the provider, which uses a single set of credentials from their config file.

It's possible that we don't want this context.project_id line at all, but I thought that maybe there's some corner case where somehow an offer is created outside of the provider or something... ?

larsks: That's what I meant about throwing an error: I don't think we ever want to default to `context.project_id` here. At least, I'm trying to think of a situation where that would make sense and I'm not coming up with one.

tzumainn: That's fair! I can take this bit out.

tzumainn: Okay, should be taken out now!

larsks:
Oops, look like there are some failing tests...

tzumainn:
Whoops - should be fixed now!

PR #112, opened 2019-07-31
Use .one_or_none() instead of .first() in db get operations

### LINKED ISSUE #111: We should be using .one (or .one_or_none) instead of .first in the database api
Many of our database api `get` operations using the `.first()` filter on queries, but this inappropriate: if the query returns more than a single result, something is wrong.

larsks:
This replaces all calls to `.first()` in db/sqlachemy/api.py with
`.one_or_none()`. This will preserve the current behavior in cases
where there is a single match or not matches, but will (correctly)
raise an exception if there are multiple matches.

Closes #111

### DIFF
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -41,7 +41,7 @@ def drop_db():
 def offer_get(marketplace_offer_id, context):
 
     return get_session().query(models.Offer).filter_by(
-            marketplace_offer_id=marketplace_offer_id).first()
+            marketplace_offer_id=marketplace_offer_id).one_or_none()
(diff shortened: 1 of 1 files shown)


tzumainn:
I think using `one` would be fine too; if we're in a function that passes in a UUID, and there's no match in the database, then I think something also went wrong. But I'm fine with this since I think @ljmcgann is adding further exception handling, and throwing a flocx_market-specific error if the query returns `None` would be fine too.

larsks:
I went with `one_or_none` because this preserves the existing behavior of the code and I was aiming for the minimum change. Raising an exception in the case of no matches would require changes to a bunch of other code.

PR #113, opened 2019-08-02
Fixed minor bug related to retrieving ocrs with arg params

fvukelic:


flocx_market/api/offer_contract_relationship.py
```diff
@@ -30,7 +30,7 @@ def get(cls, offer_contract_relationship_id=None):
             if value is not None:
                 filters[key] = value
 
-        ocrs = ocr.OfferContractRelationship.get_all(g.context, **filters)
+        ocrs = ocr.OfferContractRelationship.get_all(g.context, filters)
```
larsks: Does the fact that this made it through a previous pull request mean that we are missing a test?

fvukelic: We have the tests. It's in /tests/unit/api/test_app_offer_contract_relationship.py
the names are:
- test_get_offer_contract_relationship_3_same_statuses
- test_get_offer_contract_relationship_2_same_contractids
- test_get_offer_contract_relationship_2_same_offerids

I'm not sure why did they pass before. Maybe because we are mocking stuff? @larsks

larsks: Right, because you are mocking out `flocx_market.objects.offer_contract_relationship.OfferContractRelationship.get`, the code that would actually use the `filters` argument is never called.  Let me think about that for a bit...

larsks: Yeah, okay, I don't think there's an easy way to catch this with unit tests.

PR #116, opened 2019-08-02
Validate cost value

### LINKED ISSUE #115: We should ensure non-negative cost in offers
If an offer is submitted with a negative cost that could ultimately cause undesired behavior.

larsks:
Do not accept a negative cost value in either offer or bids.

Closes #115


larsks:
Don't merge this yet; I think I want to use `@validates` rather than a check constraint.

larsks:
Okay, all set.

PR #130, opened 2019-08-06
Remove is_admin check from ocr update

### LINKED ISSUE #126: offer_contract_relationship db update operation can't be scoped to admin
If it's scoped to admin, then the provider can't update the status.

tzumainn:
The provider needs to be able to update an ocr, but the is_admin
check prevents it from doing so.

Fixes #126

### DIFF
--- a/flocx_market/tests/unit/db/sqlalchemy/test_api.py
+++ b/flocx_market/tests/unit/db/sqlalchemy/test_api.py
@@ -499,6 +499,7 @@ def test_offer_contract_relationship_update_valid(app, db, session):
     assert check.marketplace_offer_id == offer_test_id
 
 
+@pytest.mark.skip(reason="we are currently not handling scoping correctly")
 def test_offer_contract_relationship_update_invalid_scoped(app, db, session):
     contract_data, offer_test_id = create_test_contract_data_for_ocr()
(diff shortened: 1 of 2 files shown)


flocx_market/tests/unit/db/sqlalchemy/test_api.py
```diff
@@ -499,29 +499,30 @@ def test_offer_contract_relationship_update_valid(app, db, session):
...
-        context=scoped_context,
-        offer_contract_relationship_id=ocr_id,
-        values=dict(status='testing'))
-    check = api.offer_contract_relationship_get(
-        context=admin_context,
-        offer_contract_relationship_id=ocr_id)
-
-    assert check.status != 'testing'
-    assert check.marketplace_offer_id == offer_test_id
+# TODO(tzumainn): refactor once we re-enable ocr scoping for update operation
```
larsks: Better to [mark this as skipped](http://doc.pytest.org/en/latest/skipping.html):

```
@pytest.mark.skip(reason="we are not currently handling scoping correctly")
 def test_offer_contract_relationship_update_invalid_scoped(app, db, session):
...
```

tzumainn: Ah, gotcha! Fixed.

PR #134, opened 2019-08-06
[WIP] Use an integer primary key column

### LINKED ISSUE #124: We should use a simple integer for the database primary key column
Using a UUID for the database primary key leads to wasted space (36 bytes for every foreign key reference for a UUID, vs 8 for a 64-bit integer).

larsks:
Introduce an integer primary key column for all tables. The object
UUID will continue to be the externally visible id for objects, but
using an integer primary key reduces the storage required for foreign
key references.

Closes #124

### DIFF
--- a/flocx_market/db/sqlalchemy/api.py
+++ b/flocx_market/db/sqlalchemy/api.py
@@ -221,9 +221,9 @@ def contract_create(values, context):
         contract_ref.update(values)
         contract_ref.save(get_session())
         # update foreign key for offers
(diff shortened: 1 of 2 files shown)


tzumainn:
I admit, I'd be nervous merging something like this until post-devconf!

larsks:
You'll notice the commit is marked `[WIP]` :). I've updated the PR to match. I'm concerned that this will break despite passing tests because we are handling the offer-contract-relationship table incorrectly.

PR #136, opened 2019-08-07
Remove offer server_id unique constraint

### LINKED ISSUE #50: server_id should not be unique
attempting to create an offer with the same server_id as an existing offer will fail:

```
E       sqlalchemy.exc.IntegrityError: ... UNIQUE constraint failed: offers.server_id
```

It should be possible create multiple offers from a server as long as they don't overlap in time.

tzumainn:
This PR removes the offer server_id unique constraint, and also
adds a check to ensure that a new offer for a server cannot be
created if there is an existing open offer for that server.

Fixes #50


tzumainn:
In the longer term we'd probably throw a more personal exception.

PR #141, opened 2019-08-09
Bugfix for 118 and 99

ljmcgann:
Addresses github issues 118 and 99 by checking that the types for a post
request are correct. TG-177


flocx_market/api/bid.py
```diff
@@ -20,6 +20,9 @@ def get(cls, marketplace_bid_id=None):
     def post(cls):
 
         data = request.get_json(force=True)
+        if type(data['server_config_query']) is not dict or \
+                data['server_config_query']['specs'] is None:
+            return {'message': 'Bad bid config.'}, 404
```
larsks: @tzumainn Should we be performing this check in the model (like we do for `cost` in https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/db/sqlalchemy/models.py#L45)?

flocx_market/tests/unit/api/test_app_bid.py
```diff
@@ -19,7 +19,7 @@
     end_time=now,
     duration=16400,
     status="available",
-    server_config_query={'foo': 'bar'},
+    server_config_query={'specs': 'bar'},
```
larsks: @ljmcgann I'm a little confused; shouldn't `server_config_query` be a list (of tuples)? Maybe I missed some discussion when I was on vacation or something...

flocx_market/api/bid.py
larsks: Also, shouldn't `data['server_config_query']` be a list?

tzumainn: Ah, yeah, this check should probably be done in the db model, or perhaps the db api.

`server_config_query` is apparently a dict that has a property named `specs` that specifies a list of matcher requirements.

tzumainn: (relevant matcher code: https://github.com/CCI-MOC/flocx-market/blob/master/flocx_market/matcher/match_engine.py#L25)

larsks: > Ah, yeah, this check should probably be done in the db model, or perhaps the db api.

...so, uh, which do you prefer? Previously we had discussed the db model being the right place to do data validation, but were you thinking that this situation is somehow different so that maybe the db api is a better place?

tzumainn: Oh, I just wasn't sure how to validate this within the model itself, but I'm not suuuuper experienced with this.

ljmcgann:
Is there some kind of secret to this? I tried to make a function following the same pattern as Lar's cost code, and I get an error
ERROR oslo_db.sqlalchemy.exc_filters sqlite3.IntegrityError: NOT NULL constraint failed: bids.server_config_query
Everytime. The function I make can see the value just fine but this error still gets thrown. I can change the model to make nullable true to pass but this leads to weird errors in the matcher

larsks:
@ljmcgann does your verifier function return a value?

ljmcgann:
That was it

ljmcgann:
Do we want the serv_config for an offer to also be a dictionary and have a key 'properties'?

flocx_market/db/sqlalchemy/models.py
```diff
@@ -45,7 +46,16 @@ class Bid(Base):
...
-            raise ValueError('Cost must be >= 0')
+            raise e.InvalidInput('Cost must be >= 0')
+        return value
+
+    @orm.validates('server_config_query')
+    def validate_server_config_query(self, key, value):
+        if type(value) is not dict:
+            raise e.InvalidInput('server_config_query must be a dictionary')
+        if 'specs' not in value:
+            raise e.InvalidInput("server_config_query must contain key \
```
larsks: Minor nit: you don't need `\` for extending the line here. You can just write:

```
             raise e.InvalidInput("server_config_query must contain key "
                                  "'specs'")
```

(Because multiple adjacent strings are concatentated, as in `print('this ' 'is a ' 'test')`.)

flocx_market/db/sqlalchemy/models.py
```diff
@@ -45,7 +46,16 @@ class Bid(Base):
     @orm.validates('cost')
     def validate_cost(self, key, value):
         if value < 0:
-            raise ValueError('Cost must be >= 0')
+            raise e.InvalidInput('Cost must be >= 0')
```
larsks: What's the motivation for using `InvalidInput` here rather than the built-in `ValueError` (which seems appropriate for this particular use case)?

ljmcgann: I did this because the
return json.dumps(e.message), e.code
that every function in the flask api returns requires a message and code, which ValueError does not have. Alternatively I could add an except ValueError part to all the api pieces and return the error code there, if you prefer that solution

ljmcgann:
I knew I was gonna get called out on the InvalidInput exception but as Mainn said, I didn't want to catch all ValueErrors just to throw another error with the code and message. Also that just seems worse as if a ValueError gets thrown in some unexpected part of the code we want to know about it. We know all the places where InvalidInput can be thrown so that behaviour isn't unexpected.

Issue #145, opened 2019-09-10
Offers should specify resource_type/resource_id rather than server_id


tzumainn:
Combining esi-leap into flocx-market means that flocx-market needs to be able to manipulate the specified servers. However esi-leap actually allows for a variety of node types, some for ease of testing. Updating the flocx-market data model is a prerequisite to porting that functionality here.

PR #146, opened 2019-09-10
Update data models to be non-server specific

tzumainn:
esi-leap has the capability of supporting multiple resource types;
porting esi-leap into flocx-market requires updating the data
models in flocx-market to match.

Addresses #145


flocx_market/db/sqlalchemy/api.py
```diff
@@ -60,9 +60,13 @@ def offer_get_all_by_project_id(context):
         project_id=context.project_id).all()
 
 
-def offer_get_all_by_server_id(context, server_id, status=None):
+def offer_get_all_by_resource(context,
```
larsks: I'm not sure I understand why both `resource_id` and `resource_type` are required here. Shouldn't `resource_id` be unique even across resource types? It doesn't seem like we would need to know the resource type here.

larsks: Generally: shouldn't resource type be an attribute of the resource, rather than of Offers or other objects?

tzumainn: *Is* resource_id guaranteed to be unique? Are we sure we'd only ever consider OpenStack that use a uuid as their primary identifier?

Issue #147, opened 2019-09-17
flocx-market should handle resources


tzumainn:
If multiple providers are unneeded, than esi-leap functionality should be moved to flocx-market

PR #149, opened 2019-09-20
Support resource objects in flocx-market

tzumainn:
Addresses #147

### DIFF
--- a/flocx_market/api/app.py
+++ b/flocx_market/api/app.py
@@ -30,11 +30,11 @@ def create_app(app_name):
     api.add_resource(Offer,
                      '/offer',
                      '/offer/',
-                     '/offer/<string:marketplace_offer_id>')
+                     '/offer/<string:offer_id>')
     api.add_resource(Bid,
                      '/bid',
                      '/bid/',
-                     '/bid/<string:marketplace_bid_id>')
+                     '/bid/<string:bid_id>')
     api.add_resource(Contract,
                      '/contract',
                      '/contract/',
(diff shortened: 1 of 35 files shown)


flocx_market/api/app.py
```diff
@@ -30,11 +30,11 @@ def create_app(app_name):
     api.add_resource(Offer,
                      '/offer',
                      '/offer/',
-                     '/offer/<string:marketplace_offer_id>')
+                     '/offer/<string:offer_id>')
```
larsks: :+1:

flocx_market/db/sqlalchemy/api.py
```diff
@@ -73,10 +76,10 @@ def offer_get_all_by_resource_id(context,
 def offer_get_all_unexpired(context):
     if context.is_admin:
         return get_session().query(models.Offer).filter(
-            models.Offer.status != 'expired').all()
+            models.Offer.status != statuses.EXPIRED).all()
```
larsks: :+1:

flocx_market/db/sqlalchemy/models.py
```diff
@@ -35,7 +37,8 @@ class Bid(Base):
     start_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
     end_time = orm.Column(orm.DateTime(timezone=True), nullable=False)
     duration = orm.Column(orm.Integer, nullable=False)
-    status = orm.Column(orm.String(15), nullable=False, default='available')
+    status = orm.Column(
+        orm.String(15), nullable=False, default=statuses.AVAILABLE)
```
larsks: As long as we're using status constants, why not make this an integer (or a single character) instead of a 15-character string?

tzumainn: I do think there's some value to making the status readable when querying a database.

Issue #150, opened 2019-10-04
Add oslo.policy checks


tzumainn:
This will help in alternative usages where a single admin has rights over all nodes.

PR #151, opened 2019-10-04
Add oslo.policy checks

tzumainn:
This PR adds oslo.policy checks, enabling alternative usages where
a single admin has administrative rights over all nodes.

Addresses #150

### DIFF
--- a/flocx_market/api/bid.py
+++ b/flocx_market/api/bid.py
@@ -1,25 +1,31 @@
 from flask_restful import Resource
 from flask import request, g
+import json
+
 from flocx_market.objects import bid
 from flocx_market.common import exception
-import json
+from flocx_market.common import policy
 
 
 class Bid(Resource):
 
     @classmethod
     def get(cls, bid_id=None):
+        cdict = g.context.to_policy_values()
 
         if bid_id is None:
+            policy.authorize('flocx_market:bid:get_all', cdict, cdict)
             return [x.to_dict() for x in bid.Bid.get_all(g.context)]
-
         try:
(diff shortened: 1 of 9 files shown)


flocx_market/api/bid.py
```diff
@@ -1,25 +1,31 @@
 from flask_restful import Resource
 from flask import request, g
```
larsks: Really, flask? You're going to name a public API attribute `g`?

flocx_market/common/policy.py
```diff
@@ -0,0 +1,165 @@
...
+CONF = flocx_market.conf.CONF
+_ENFORCER = None
+
+default_policies = [
+    policy.RuleDefault('is_admin',
+                       'role:admin or role:flocx_market_admin',
+                       description='Full API access'),
+]
+
+bid_policies = [
```
larsks: If I'm reading this properly, the default policies are all admin-only, right? The idea being that an operator would always need to deploy a customized policy file to make the service generally available?

flocx_market/api/bid.py
tzumainn: I had a very similar reaction :)

flocx_market/common/policy.py
tzumainn: Yes. I have decided this is the best way to preserve backwards compatibility with everyone using this code.

larsks: I don't think there *is* anyone using this code :)

tzumainn: what, no, I'm sure there are billions

